PR #9, opened 2014-10-20
jep exceptions include full python traceback and/or java stacktrace

ndjensen:
This changeset makes JepExceptions include the entire Python traceback and Java Stacktrace, depending on where the original error was thrown.

This perhaps need to be more configurable as some applications may not want to include either trace while other applications will require full traces to properly diagnose and fix bugs.

### DIFF
--- a/src/jep/Util.java
+++ b/src/jep/Util.java
@@ -1,5 +1,8 @@
 package jep;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
 /**
  * <pre>
  * Util.java - utility functions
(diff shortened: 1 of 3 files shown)


ndjensen:
Agreed that it is not ideal and we should have the JepException contain the original JavaException, probably just through the constructor JepException(Throwable).

For our purposes we'd like the python traceback associated with the JepException though, complete with traceback with python calls and line numbers.  That all gets logged so we can fully trace what happened with an error in our production environment.  I don't see how to do that if we use the jep module?

Regardless I am ok with dropping this pull request until we figure out a more elegant solution.

PR #20, opened 2015-08-06
Fix #19 patch build scripts to work on Mac OSX (Yosemite)

### LINKED ISSUE #19: MacOSX build doesn't work with default Python 3.4 distribution
Under MacOSX I clone the jep repository.
Download and install package for Mac OSX 32/64bit from python web site.
Download and install Oracle java 7 from oracle site.
When I run python3 setup.py build it fails with
```
$ python3 setup.py clean
Traceback (most recent call last):
  File "setup.py", line 100, in <module>
    include_dirs=get_java_include() + ['src/jep', 'build/include'] + numpy_include,
  File "/Users/builder/tmop/jep/commands/java.py", line 64, in get_java_include
    inc = os.path.join(get_java_home(), inc_name)
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/posixpath.py", line 89, in join
    "components") from None
TypeError: Can't mix strings and bytes in path components'
```
If you fix java.py script then build breaks again with
> ld: library not found for -lpython3.4m

Fixed python.py script I gain the native lib for mac os :8ball:

nfalco79:


ndjensen:
Thanks for testing this and providing the fix!

PR #37, opened 2015-12-16
Use sequence protocol to support FrameSummary objects

thecom:
Jep will crash the JVM with Python 3.5 in case of an exception otherwise.

I've not checked older Python versions or run the tests. I don't think this change causes trouble because as far as I known tuples implement the sequence protocol, too.


ndjensen:
I agree that fix should be backwards compatible and safe.  Thanks for the fix!

ndjensen:
Merged into master and then merged into dev_3.4 and dev_3.5 branches.

PR #50, opened 2016-05-09
Support setting python initialization variables

jvstein:
- Separated native library loading from initialization of python interpreter.
- Cleaned up UnsatisfiedLinkError stacktrace.
- Added Jep.setInitParams method to be callable before creating a Jep instance.

### DIFF
--- a/src/jep/Jep.java
+++ b/src/jep/Jep.java
@@ -101,7 +101,7 @@ public final class Jep implements Closeable {
      * be created from.
      */
     private static class TopInterpreter implements Closeable {
-        Object initLock;
+        Object initLock = new Object();
         Throwable error;
 
         /**
@@ -117,23 +117,22 @@ public final class Jep implements Closeable {
 
                 @Override
                 public void run() {
-                    try {
-                        initializePython();
-                    } catch (Throwable t) {
-                        error = t;
-                    } finally {
-                        synchronized (TopInterpreter.this) {
-                            TopInterpreter.this.notify();
+                    synchronized(initLock) {
+                        try {
+                            initializePython();
+                        } catch (Throwable t) {
+                            error = t;
+                        } finally {
+                            synchronized (TopInterpreter.this) {
+                                TopInterpreter.this.notify();
+                            }
                         }
-                    }
-                    /*
-                     * We need to keep this top interpreter thread around. It's
-                     * not going to be used again but if its thread shuts down
-                     * while another thread is in python, then the thread state
-                     * can get messed up leading to stability/GIL issues.
-                     */
-                    initLock = new Object();
-                    synchronized (initLock) {
+                        /*
+                         * We need to keep this top interpreter thread around. It's
+                         * not going to be used again but if its thread shuts down
+                         * while another thread is in python, then the thread state
+                         * can get messed up leading to stability/GIL issues.
+                         */
                         try {
                             initLock.wait();
                         } catch (InterruptedException e) {
@@ -162,10 +161,9 @@ public final class Jep implements Closeable {
          * Stop the interpreter thread.
          */
         @Override
-        public synchronized void close() {
-            if (null != initLock) {
+        public void close() {
+            synchronized(initLock) {
                 initLock.notify();
-                initLock = null;
             }
         }
     }
--- a/src/jep/Jep.java
+++ b/src/jep/Jep.java
@@ -57,6 +57,8 @@ public final class Jep implements Closeable {
 
     private static final String THREAD_WARN = "JEP THREAD WARNING: ";
 
+    private static TopInterpreter topInterpreter = null;
+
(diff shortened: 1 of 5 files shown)


ndjensen:
Need a few days for me and Ben to digest all this.  Your enthusiasm is awesome but we can't keep up given the non-Jep stuff we have to do.

src/jep/Jep.java
```diff
@@ -129,7 +132,7 @@ public void run() {
                      * while another thread is in python, then the thread state
                      * can get messed up leading to stability/GIL issues.
                      */
```
bsteffensmeier: It is theoretically possible to call call close() here, before the initLock is created. Once you notify TopInterpreter,this there is nothing to prevent that other thread from running long enough to call close() before running this thread any further. I think the simplest solution may be to create initLock earlier and just synchronize the entire method on initLock.

jvstein: Oh, that's a good point.

Thanks for bearing with my bad threading code. It's been a couple years since I've done any non-trivial threading and that was all in C# and never with native modules.

jvstein:
Absolutely, and sorry for the barrage. I had an immediate itch to scratch and some time to devote to it.

src/jep/Jep.java
jvstein: I think I fixed the issue with the zombie thread. I tested it with `Thread.sleep` calls in a couple different locations and it's all executing in the desired order, with the interpreter thread stopping nicely.

For what it's worth, a colleague mentioned that using a `java.util.concurrent.CountDownLatch` or some other abstraction might be a little clearer.

ndjensen:
I'm planning on merging this in the near future.  Just two questions to make sure we're on the same page.

1. Are there any unit tests?  i.e. Unit tests as in unit tests we can run with `python setup.py test`, or were you just testing with your local application(s)?
1. After this merged are you considering it done?  If so that's fine, but then I will go forward and analyze the other flags to see if they'd be useful and proceed onward.  And I need to make it use a default PyConfig if none is provided for reasons as stated in the ticket.

jvstein:
1. There are no unit tests that cover these scenarios. I wrote some tests that cover my specific configuration in my local application (built/tested with maven). It looks like the existing unit tests are executed with the `jep` script and all in the same process. I'd need to write the unit tests in java and execute each test in a separate process. I'll see what I can do to move my existing test cases over to the python.
2. I'm done for the immediate term. I'll focus on getting unit tests written first if you want to look into other flags and get the config working more nicely.

ndjensen:
Yeah the jep script will launch Jep itself, so all the tests that have to start a Jep instance aren't  hooked into the unittest framework at present.  You can see in the jep.test package that in some cases I just wrote some Java mains(), they're not hooked into the unittest framework.  At some point that needs to be solved but that's not your problem to solve that.  So you could go my cheater route of a Java main() that won't get exercised regularly but only sometimes, or wait for us to set up the tests better.  One of the goals of dev_3.6, presuming we achieve it, is to improve the unittests.

Ok I will take care of the rest of what we discussed.  Not sure when I will get to it.

ndjensen:
Ok, talking with my coworkers here, in situations where we need to launch a unique Jep instance for the test, we should establish a pattern of using subprocess.popen(args...) and listen to stdout for something, then assert that the stdout matches what we were expecting.

For example, I just added [this file](https://github.com/mrj0/jep/blob/dev_3.6/src/jep/test/TestNoneDictionary.java) which was provided by the ticket reporter on [ticket 40](https://github.com/mrj0/jep/issues/40) and slightly modified by me.  The quick test was to just look at the system output, but really that should be run every time the unit tests run to ensure we don't regress.  But since it requires a new Jep instance (and you can't safely create those on the same thread) to call getValue(), there wasn't a good way to hook into Python's unittest framework.

Listening for stdout and asserting seems like the best approach forward for these kinds of tests and then preventing regressions.

ndjensen:
I'm going to get someone here to work on providing an example and possibly a framework for doing the popen() style separate process tests using the aforementioned Java test and ticket.  Cause we really need to prevent regressions and provide more confidence that the tests are exercising enough functionality.

ttgurney:
Here's a tiny module I threw together for the above-mentioned pattern:
```python
import subprocess
from contextlib import contextmanager

@contextmanager
def jep_pipe(argv):
    if isinstance(argv, str):
        argv = [argv]
    try:
        p = subprocess.Popen(['jep'] + argv, stdout=subprocess.PIPE)
        yield p.stdout
    finally:
        p.kill()
```

And say you want to test the behavior of your program `hello.py` run inside Jep:
```python
print 'Hello, world!'
```
So with this, in a test you can do:
```python
def test_jep_hello():
    with jep_pipe('hello.py') as p:
        assert p.readline() == 'Hello, world!\n'
```

Issue #52, opened 2016-06-27
NDArray needs to support unsigned dtypes


tverbele:
I noticed that 
```
jep.eval("t = somePythonMethodThatReturnsNDarray()");
Object t = jep.getValue("t");
```
results in an Exception
```
jep.JepException: jep.JepException: <type 'exceptions.TypeError'>: Unable to determine corresponding Java type for ndarray
```

I would have expected that this would result into an NDArray object in Java. Is there an other/better way to fetch NDArray data from Python in Java?

ndjensen:
That's supposed to work.  What version of Jep are you using?  Do you know what the dtype is of the underlying ndarray?  At present we only support dtypes that cross over to java well (int8, int16, int32, int64, float32, float64).  We might need a better error message if the dtype is the underlying cause.

tverbele:
I looked further into this. Indeed it is because the datatype appeared to be unsigned bytes instead of bytes.

ndjensen:
Thank you for contributing the commit!  We're going to want this on the dev_3.6 branch instead of master.  We're possibly different than other projects, but at present master is the released version and the dev branches are for new code.  I need to make that clearer in the CONTRIBUTING.md file.  @bsteffensmeier or I can probably rebase it against dev_3.6 if you don't want to do that.

@bsteffensmeier, can you review the [commit referenced above](https://github.com/mrj0/jep/pull/53/commits/a8f7b42748c129216c2346ece803b31ab2330b2a)?  It looks good to me but it's going to conflict with some of your dev_3.6 work for numpy, in particular the speedup you added and the Java char[] <-> npy_uint16.

We're also going to need to write some tests for unsigned data.

ndjensen:
And we should strive to make the error message point out the dtype if that's possible, such as

"Unable to determine corresponding Java type for ndarray **of dtype ____**"

bsteffensmeier:
I had a few minor comments inline but nothing major

It is going to conflict terribly with the numpy changes I have on dev_3.6. The optimizations should be a straightforward merge but the fact that I converted UINT16 to a char[] is incompatible with this. Since this gives us support for all the unsigned types I think we should remove the char[] conversion and use this instead to keep it consistent. When we remove support for char[] we need to make sure we throw an exception from the java constructor if anyone tries.

tverbele:
I'm sorry for committing on top of the wrong branch. Will you guys merge further in dev_3.6 or should I do a new PR?

[...]

PR #53, opened 2016-06-30
support unsigned data in NDArrays

tverbele:
I added support for the unsigned variants of BYTE, INT16, INT32 and INT64 to be converted to a Java NDArray. To reflect this in Java, I added an unsigned flag in the NDArray class to indicate one has to treat the data as unsigned. 

This resolves the issue I reported #52


src/jep/jep_numpy.c
```diff
@@ -408,6 +429,16 @@ PyObject* convert_jndarray_pyndarray(JNIEnv *env, jobject obj)
...
 
+    if (ndarrayIsUnsigned == 0) {
+        ndarrayIsUnsigned = (*env)->GetMethodID(env, JEP_NDARRAY_TYPE, "isUnsigned",
+                                             "()Z");
+        if (process_java_exception(env) || !ndarrayIsUnsigned) {
+            return NULL;
+        }
+    }
+
+    usigned = (*env)->CallBooleanMethod(env, obj, ndarrayIsUnsigned);
```
bsteffensmeier: Need to process the java exception to ensure the method didn't throw anything.

src/jep/jep_numpy.c
```diff
@@ -141,28 +142,6 @@ jarray convert_pyndarray_jprimitivearray(JNIEnv* env,
     sz = (jsize) PyArray_Size(param);
     paType = PyArray_TYPE((PyArrayObject *) param);
 
-    if (desiredType == NULL) {
```
bsteffensmeier: The function documentation should be updated since NULL is no longer valid.

PR #67, opened 2017-01-18
Convert python callables to functional interfaces.

Techcable:
A similar conversion is implemented in nashorn, so I figured it'd be good here too.

### DIFF
--- a/src/jep/InvocationHandler.java
+++ b/src/jep/InvocationHandler.java
@@ -6,31 +6,32 @@
  * This software is provided 'as-is', without any express or implied
  * warranty. In no event will the authors be held liable for any
  * damages arising from the use of this software.
- * 
+ *
  * Permission is granted to anyone to use this software for any
  * purpose, including commercial applications, and to alter it and
  * redistribute it freely, subject to the following restrictions:
- * 
+ *
  *     1. The origin of this software must not be misrepresented; you
  *     must not claim that you wrote the original software. If you use
  *     this software in a product, an acknowledgment in the product
  *     documentation would be appreciated but is not required.
- * 
+ *
  *     2. Altered source versions must be plainly marked as such, and
  *     must not be misrepresented as being the original software.
- * 
+ *
  *     3. This notice may not be removed or altered from any source
  *     distribution.
  */
 package jep;
 
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 import jep.python.PyObject;
 
 /**
  * Handle Proxy method calls.
- * 
+ *
  * @author [mrjohnson0 at sourceforge.net] Mike Johnson
  */
 public class InvocationHandler implements java.lang.reflect.InvocationHandler {
@@ -40,11 +41,11 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
     private long target;
 
     private Jep jep;
-    private final boolean isDirect;
+    private final boolean functionalInterface;
 
     /**
      * Creates a new <code>InvocationHandler</code> instance.
-     * 
+     *
      * @param tstate
      *            the thread state id
      * @param ltarget
@@ -54,12 +55,12 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
      * @exception JepException
      *                if an error occurs
      */
-    public InvocationHandler(long tstate, long ltarget, Jep jep, final boolean isDirect)
+    public InvocationHandler(long tstate, long ltarget, Jep jep, final boolean functionalInterface)
             throws JepException {
         this.tstate = tstate;
         this.target = ltarget;
         this.jep = jep;
-        this.isDirect = isDirect;
+        this.functionalInterface = functionalInterface;
 
         // track target with jep.
 
@@ -76,7 +77,7 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
      * Processes a method invocation on a proxy instance and returns the result.
      * This method will be invoked on an invocation handler when a method is
      * invoked on a proxy instance that it is associated with.
-     * 
+     *
      * @param proxy
      *            the proxy instance that the method was invoked on
      * @param method
@@ -121,6 +122,14 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
             throws Throwable {
         this.jep.isValidThread();
 
+        if (this.functionalInterface && (method.getModifiers() & Modifier.ABSTRACT) == 0) {
+            /*
+             * If it's a non-abstract (default) interface method,
+             * fallback to invoking the default implementation if we're a functional interface.
+             */
+            return method.invoke(proxy, args);
+        }
+
         // java passes null args sometimes. *shrugs*
         if (args == null)
             args = new Object[0];
@@ -132,9 +141,9 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
             types[i] = Util.getTypeId(args[i]);
 
         return invoke(method.getName(), this.tstate, this.target, args, types,
-                Util.getTypeId(method.getReturnType()), this.isDirect);
+                Util.getTypeId(method.getReturnType()), this.functionalInterface);
     }
 
     private static native Object invoke(String name, long tstate, long target,
-            Object[] args, int[] types, int returnType, boolean isDirect);
+            Object[] args, int[] types, int returnType, boolean functionalInterface);
 }
--- a/src/jep/InvocationHandler.java
+++ b/src/jep/InvocationHandler.java
@@ -40,6 +40,7 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
     private long target;
 
     private Jep jep;
+    private final boolean isDirect;
 
     /**
      * Creates a new <code>InvocationHandler</code> instance.
@@ -53,11 +54,12 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
      * @exception JepException
      *                if an error occurs
      */
-    public InvocationHandler(long tstate, long ltarget, Jep jep)
+    public InvocationHandler(long tstate, long ltarget, Jep jep, final boolean isDirect)
             throws JepException {
         this.tstate = tstate;
         this.target = ltarget;
         this.jep = jep;
+        this.isDirect = isDirect;
 
         // track target with jep.
 
@@ -130,9 +132,9 @@ public class InvocationHandler implements java.lang.reflect.InvocationHandler {
             types[i] = Util.getTypeId(args[i]);
 
         return invoke(method.getName(), this.tstate, this.target, args, types,
-                Util.getTypeId(method.getReturnType()));
+                Util.getTypeId(method.getReturnType()), this.isDirect);
     }
 
     private static native Object invoke(String name, long tstate, long target,
-            Object[] args, int[] types, int returnType);
+            Object[] args, int[] types, int returnType, boolean isDirect);
 }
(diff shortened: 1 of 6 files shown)


src/jep/convert_p2j.c
```diff
@@ -630,6 +735,8 @@ jobject PyObject_As_jobject(JNIEnv *env, PyObject *pyobject,
         return pyfastsequence_as_jobject(env, pyobject, expectedType);
     } else if (PyDict_Check(pyobject)) {
         return pydict_as_jobject(env, pyobject, expectedType);
+    } else if (PyCallable_Check(pyobject) && isFunctionalInterfaceType(env, expectedType)) {
```
bsteffensmeier: isFunctionalInterfaceType has alot of error conditions that return NULL,  we need error handling here also to ensure it propagates.

bsteffensmeier:
1. I think we have problems with default methods and the proxy objects. I haven't played with it at all yet, but based off [what I found online](https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/) the current proxy object completely ignores default method implementations. With your changes default method invocations will be forward through the proxy and execute the callable directly which seems to make our lack of default method handling much worse.
1. Unfortunately, in order to support python 2.7 on windows we must use the MSVC 2008 compiler which does not support C99. I know for a fact that you must move all your variable declarations to the top of your functions and I am guessing you can't use stdbool.h. It pains me to support these limitations but I don't think we want to drop support and there is no other way. Ive started using -Wdeclaration-after-statement in gcc since Ive checked in "bad" code more than once.
1. Technically we still support java back to 1.6. The only thing I see in your code that is specific to 1.8 is the tests. Can you write a test that is compatible with 1.6 and then skip the other tests if the java is too old?

src/jep/convert_p2j.c
```diff
@@ -41,6 +42,9 @@
 
 #define JCHAR_MAX   0xFFFF
 
+static jmethodID classGetMethods = 0;
+static jmethodID classGetModifiers = 0;
+static jmethodID methodGetModifiers = 0;
```
ndjensen: We've got `classGetMethods`  in pyjobject.c, `classGetModifiers` in pyjclass.c, and `methodGetModifiers` in pyjmethod.c.  We need to consolidate to....somewhere.  Now that I look at it we have `Class.getComponentType()` in jep_util.c and pyjarray.c too.  `Class.getParameterTypes()` is in pyjconstructor.c and pyjmethod.c.

Doesn't need to be fixed here.  @bsteffensmeier, we should try to consolidate those.  Maybe we rip more out of jep_util.c and make a jep_init.c or jep_cache.c or jep_reflection.c.  It's not a huge hit since we cache the jmethodIDs but it would still be nice to have the reusable ones in one location.

src/jep/convert_p2j.c
```diff
@@ -52,6 +56,58 @@ static jmethodID unmodifiableList      = 0;
...
+    if (!JNI_METHOD(methodGetModifiers, env, JMETHOD_TYPE, "getModifiers", "()I")) {
+        process_java_exception(env);
+        return NULL;
+    }
+    if ((*env)->PushLocalFrame(env, JLOCAL_REFS) != 0) {
+        process_java_exception(env);
+        return NULL;
+    }
+    int classModifiers = (*env)->CallIntMethod(env, type, classGetModifiers);
+    if (!(classModifiers & 512)) {
```
ndjensen: Any reason to not use `Class.isInterface()` here instead of `Class.getModifiers()` and comparing the modifier value?

src/jep/convert_p2j.c
```diff
@@ -52,6 +56,58 @@ static jmethodID unmodifiableList      = 0;
...
+            return NULL;
+        }
+        if (modifiers & 1024) { // check if the method is abstract
+            // We found two different abstract methods, so we're not a functional interfaces
+            if (abstractMethod != NULL) {
+                (*env)->PopLocalFrame(env, NULL);
+                return NULL;
+            } else {
+                abstractMethod = method;
+            }
```
ndjensen: Comment of "We found two different abstract methods..." should go inside `if (abstractMethod != NULL)`.  I was confused first time reading it.

tests/test_call.py
```diff
@@ -5,6 +5,7 @@
 Boolean = jep.findClass('java.lang.Boolean')
 StringBuilder = jep.findClass('java.lang.StringBuilder')
 ArrayList = jep.findClass('java.util.ArrayList')
+LongStream = jep.findClass('java.util.stream.LongStream')
```
ndjensen: We want to maintain Java 1.6 compatibility for as long as possible.  Can we test with something else?

src/jep/InvocationHandler.java
```diff
@@ -40,6 +40,7 @@
     private long target;
 
     private Jep jep;
+    private final boolean isDirect;
```
ndjensen: I am a little lost as to what isDirect is referring to.  My head is stuck on direct memory.  Please explain.

src/jep/convert_p2j.c
bsteffensmeier: I've actually got a [commit saved off](https://github.com/bsteffensmeier/jep/commit/6081a07d7f7c2c071b2849180ceb56b74e8695ac) that I intend to merge into 3.7 that consolidates all of the java method calls. I just haven't gotten around to merging it yet. So for this review I recommend we just leave all the duplicates and I will merge my change after this goes in and clean up this with everything else.

src/jep/InvocationHandler.java
Techcable: isDirect refers to whether or not the InvocationHandler is directly invoking the object via its __invoke__ method, or invoking the corresponding.
When a normal InvocationHandler receives a call to Runnable.run(), it'd call pythonObject.run(), but if InvocationHandler.isDirect, it'd call pythonObject() so it can be used for callable objects.

ndjensen: Thank you for the explanation.  I'd like to not have "is" in the variable name, Java standard is to have the getter for booleans be isVariableName().  Since direct wasn't immediately clear to me, a better name could be functionalInterface (and getter isFunctionalInterface()).  Since that's essentially what is happening here even if we're not supporting Java 8 language features yet.

Techcable:
I've fixed the tests for java 6, removed stdbool.h, and cleaned up the issues in the code.

src/jep/InvocationHandler.java
```diff
@@ -53,11 +55,12 @@
      * @exception JepException
      *                if an error occurs
      */
-    public InvocationHandler(long tstate, long ltarget, Jep jep)
+    public InvocationHandler(long tstate, long ltarget, Jep jep, final boolean functionalInterface)
```
ndjensen: We should keep the old method signature for backwards compatibility.

```java
public InvocationHandler(long tstate, long ltarget, Jep jep) {
    this(tstate, ltarget, jep, false);
}
```
Honestly @mrj0 is most familiar with the `Proxy` and `InvocationHandler` code.  Do people new up an `InvocationHandler` or just use `Proxy.newProxyInstance(...)`?  Either we should keep the backwards compatibility as specified above or we should make the constructor protected if it's not supposed to be public API.

src/jep/Proxy.java
```diff
@@ -76,12 +80,15 @@ protected Proxy(InvocationHandler h) {
...
+            ClassLoader loader, String[] interfaces) {
+        return newProxyInstance(tstate, ltarget, jep, loader, interfaces, false);
+    }
 
+    private static Object newProxyInstance(long tstate, long ltarget, Jep jep,
+            ClassLoader loader, String[] interfaces, boolean isDirect) {
         InvocationHandler ih = null;
         try {
-            ih = new jep.InvocationHandler(tstate, ltarget, jep);
+            ih = new jep.InvocationHandler(tstate, ltarget, jep, isDirect);
```
ndjensen: I would prefer to keep using the term functionalInterface over isDirect.

src/jep/convert_p2j.c
```diff
@@ -52,6 +55,57 @@ static jmethodID unmodifiableList      = 0;
...
+        return 0;
+    }
+    if ((*env)->PushLocalFrame(env, JLOCAL_REFS) != 0) {
+        process_java_exception(env);
+        return 0;
+    }
+    if (!(*env)->CallBooleanMethod(env, type, classIsInterfaceMethod)) {
+        return 0; // It's not an interface, so it can't be functional
+    }
+    jobjectArray methods = (jobjectArray) (*env)->CallObjectMethod(env, type, classGetMethods);
```
ndjensen: You have to declare all the variables at the top of the block.  It's annoying but required for some compilers.

src/jep/InvocationHandler.java
```diff
@@ -119,6 +122,14 @@ public Object invoke(Object proxy, Method method, Object[] args)
             throws Throwable {
         this.jep.isValidThread();
 
+        if (this.functionalInterface && (method.getModifiers() & Modifier.ABSTRACT) == 0) {
+            /*
+             * If it's a non-abstract (default) interface method,
+             * fallback to invoking the default implementation if we're a functional interface.
+             */
+            return method.invoke(proxy, args);
```
bsteffensmeier: This will call itself, I wrote a small test class to verify and I got a StackOverflowError

```java
public class Test{

    public static void main(String[] args) throws Throwable{
        try(Jep jep = new Jep()){
            jep.set("test", new Test());
            jep.eval("test.checkConsumer(lambda: True)");
        }
    }

    public void checkConsumer(Consumer<?> consumer){
        consumer.andThen(null);
    }

}
```

src/jep/convert_p2j.c
```diff
@@ -52,6 +55,57 @@ static jmethodID unmodifiableList      = 0;
...
+    if (process_java_exception(env)) {
+        (*env)->PopLocalFrame(env, NULL);
+        return 0;
+    }
+    jsize numMethods = (*env)->GetArrayLength(env, methods);
+    jobject abstractMethod = NULL;
+    for (jsize i = 0; i < numMethods; i++) {
```
bsteffensmeier: this fails to compile on my linux/gcc:

src/jep/convert_p2j.c:85: error: ‘for’ loop initial declarations are only allowed in C99 mode

I would like to enable c99 but python 2.7/msvc is holding us back.

tests/test_call.py
```diff
@@ -89,3 +88,32 @@ def test_overload(self):
...
+            .filter(lambda i: i % 2 == 0)\
+            .reduce(lambda first, second: first + second)\
+            .getAsLong()
+        self.assertTrue(result == sum(range(2, 1000, 2)))
+
+    def test_observer(self):
+        from java.util.concurrent import Executors
+        hasNachos = False
+        def runTask():
+            nonlocal hasNachos
```
bsteffensmeier: This test fails on python 2.7 because nonlocal was introduced in python 3

src/jep/InvocationHandler.java
bsteffensmeier: I've done more investigation into handling default methods. The generally accepted solution is to use MethodHandles, however that is not available until java 1.7. I think we could write some JNI that uses CallNonvirtual<type>Method to accomplish the same thing but it would be an annoyingly large amount of code. It looks like this would make a good excuse to drop support of java 1.6.

Since its getting complicated and default method handling is already somewhat broken anyway I think we should not try to fix it with this pull request. I do think we should try to fail more gracefully than a StackOverflowError, the best option I see is to just use invoke with functionalInterfacec=false, which will probably end up with an Attribute Not Found type of error. Those errors would be slightly easier to identify, and if we don't solve default methods it would also give developers the ability to reimplement the default method in python to avoid errors.

src/jep/InvocationHandler.java
mrj0: I think most people would use Proxy.. but maybe people are using it with their own implementations.

PR #88, opened 2017-08-13
README.rst: Update supported python versions

dbaio:
Follow 332062d3c2a0e034e6c063b14c1e105497d9ed01


ndjensen:
Thanks for catching that.  Unfortunately I can't upload a new one to pypi.org without incrementing the version so until there is a 3.7.1 it will remain without Python 3.6 listed there.

Issue #35, opened 2015-12-08
Switch to maven?


Techcable:
Could you please switch to maven to allow us to embed jep in our application?
It'd be really nice to be able to embed jep in our applications, instead of requiring it to be on the path.

The python code can be included in the jar by simply placing it in a custom directory, like 'src/main/python', and telling python to look there.
As for the native code [this](https://github.com/maven-nar/nar-maven-plugin) looks interesting.

ndjensen:
This is hardly the first time this has been requested so I'll leave the ticket open.  But long story short, I don't want to get in the business of building and packaging the compiled native libraries and no one else has volunteered.  The [maven-nar](https://github.com/maven-nar/nar-maven-plugin) package you linked to looks cool but it appears to only account for "machine architecture, operating system, and linker".  When compiling the jep native library, it needs to be built with the intended:
* operating system
* Python version
* numpy support enabled or not
  * if enabled, numpy version

That suddenly results in a lot of built libraries if we want to try and support everyone's needs.  And something like maven-nar doesn't account for specifying Python version or numpy version.  Even [Python's binary distribution format](https://www.python.org/dev/peps/pep-0425/) doesn't allow us to distinguish to the granularity necessary.  Maven support is not a priority at this time, I don't see a good way (yet) to support it without getting bogged down by attempting to distribute the compiled native code.

ndjensen:
> It'd be really nice to be able to embed jep in our applications, instead of requiring it to be on the path.

Would it be helpful/useful if Jep had the Java src and jar(s) available through maven but didn't provide the native library through that route?  Thinking about it more, a developer using Jep had to build and/or install Python (and optionally numpy) and they didn't accomplish that through maven.  Maybe we're biting off more than we can chew by trying to solve the native problem, but having maven support for only the Java side would be more achievable.

Techcable:
Just having the api as a maven artifact to specify as a dependency would be very nice.
Something we could specify as `<scope>provided</scope>` in our maven projects, while requiring it to be on the classpath.

Another option would be replacing the JNI bindings with one of the better ffi libraries for java.
Something like [BridJ](https://github.com/nativelibs4java/BridJ), [Java Native Runtime](https://github.com/jnr), or [Java Native Access](https://github.com/java-native-access/jna).
It would solve the redistribution problem and allow you to easily switch to maven, but would require a port of the C portion (which is most of the project).

ndjensen:
[...]

Issue #29, opened 2015-09-07
Java packages starting with io can conflict with Python's io module


felixnext:
Hey,
I am using Jep 3.4.1 (installed with `python setup.py install`) on a server running python 2.7. I am calling Jep from Scala code. I am able to instanciate python, however I get the following error when I try to load a script:
```jep.JepException: <type 'exceptions.ImportError'>: /usr/lib/python2.7/lib-dynload/_multiprocessing.so: undefined symbol: _Py_ZeroStruct```

The script itself includes the following code:
```python
from gensim.models import doc2vec
import nltk
import string
import os
doc2vec_model = None
```
The script works if I load it in python in the console. I made some experiments and the script runs fine, if I delete the import statements. Any ideas, why that might be?

Greetings

SourceSink:
Also I just tested the jep console, where the import statements work without any problems...

SourceSink:
I also just tested to set the LD_PRELOAD variable using `export LD_PRELOAD="/usr/lib/libpython2.7.so.1.0"`, which brought me (hopefully) one step closer to the solution, as the error now is:
```
jep.JepException: <type 'exceptions.ImportError'>: io.BufferedIOBase
```
Which looks like it is related to the Java-Version involved...

ndjensen:
Hi, you've made good progress on this.  Have you checked the LD_LIBRARY_PATH environment variable?  Besides LD_PRELOAD, the jep console script sets LD_LIBRARY_PATH.  It looks like maybe you have more than one version of python on your machine?

SourceSink:
Which path goes into the LD_LIBRARY_PATH?

ndjensen:
Look at the jep script that the build produced, it should be in there.  Typically it needs to point at the python/lib dir, wherever libpython2.7.so is.

SourceSink:
Alright, thanks for the quick reply. I tried it by setting:
```
export LD_LIBRARY_PATH="/usr/lib:/usr/local/lib/python2.7/dist-packages/"
```
However I still get the same error. (Also used -Djava.library.path with no result)
Although the error seems to be quite strange, as I can import most packages except the nltk.
(More specifically: I am unable to call `from io import BufferedIOBase`. However `import io` works...).

ndjensen:
Just to make sure I'm following this correctly.

1. nltk imports correctly from the python console?
2. nltk imports correctly from the jep console?
3. nltk doesn't import when loading from a java application?

[...]

PR #90, opened 2017-08-27
add local maven repo support

ananthc:
This pull request addresses a part of the discussion given here https://github.com/ninia/jep/issues/35. 

This enables a developer to atleast run a "mvn clean install" at the root of the project to deploy the jar to the developers local repository. 

Deploying to the maven central requires additional constructs and I can create a separate pull request once the JIRA ticket is approved. 

This pull request will just simplify the local deploy process to a users local repo.


ananthc:
Hello Nate,

Thanks for the review comments.

Replies inline.


Regards,
Ananth
> On 30 Aug 2017, at 12:42 am, Nate Jensen <notifications@github.com> wrote:
> 
> Thanks for the pull request, especially given my lack of knowledge of maven. A couple of questions:
> 
> What is the jetbrains reference? Is that normal or does that add an extra dependency?
[Ananth]>> The files with the iml extension are the IDE files. I can delete them if the policy is to not include IDE specific files. Folks who would use the IntellijIdea would have an IDE file as part of the repo but it is not mandatory. Please let me know if you think that it is not a practice to include IDE files in the project.
> My hope was to name the project org.ninia.jep. Granted that does not match the current package names, but I can't change the package names without breaking backwards compatibility. But central maven already has a jep (Java Expression Parser) so we wanted to distinguish ourselves from them. Is that possible to identify ourselves with an org name like other open source projects?
[Ananth]>> Good idea. Changed the group id of the artefact to be “org.ninia.jep”. As backward compatibility might not allow us for jet, I guess this is the better alternative. It should not break anything if the group id differs from a given package path in the source code.

Techcable:
Don't you have to own the domain to put it on maven? I know you could resort to a github domain if you don't already have one.

bsteffensmeier:
I've got a few questions, I am not very familiar with maven, so some of these might be obvious:

1. Why are there multiple pom.xml files? I thought most projects only need 1.
2. What does jep-parent mean? For the artifactId think this is the name of the jar in which case I would prefer just jep. I'm not sure what it means in the name of the iml file though?
3. Are the test classes included in the jars? Ideally the test classes would be omitted from the distributed jars.
4. Is the commented out block at the bottom of src/pom.xml significant or was that just for experimentation?

This also impacts the release process, mostly I am just worried we are going to forget to update the version in the pom.xml for new releases. This would be less likely if we can have a single pom.xml but we might want to consider other ways to ensure they stay in sync. Perhaps setup.py should update the version in pom.xml. This would allow a developer to at least notice the change when looking at the git status. Another option would be to hook up a test of some sort for travis so that travis fails if the versions become out of sync. Thinking of travis, it would be good to try to hook up a maven build in travis to make sure we don't break this. It would be even better if we could automatically prove that setup.py and maven are producing identical jars. These things are all beyond the scope of this pull request, really just thinking out loud on how we can make sure this functionality remains solid.

ananthc:
@Techcable  - Ownership of the domain is not a mandatory check by maven. The group ID is encouraged to be a unique piece of string and a domain is a good and standard way to ensure that. 

The issue I see with github being part of the groupid as it "dilutes" the domain concept as github is more of a hosting site as opposed to being the "owner". Hence I think the proposal for org.ninia.jep makes more sense.

There are many references in the maven central which do not follow this convention ( ex: log4j ). Maven naming conventions guide recommends to use the domain name as a good way to ensure uniqueness. Some interesting links in this this context here : 
- https://maven.apache.org/guides/mini/guide-naming-conventions.html 
- https://stackoverflow.com/questions/19924980/groupid-of-maven-projects-hosted-on-github
- https://stackoverflow.com/questions/3150003/naming-convention-for-maven-artifacts

ananthc:
@bsteffensmeier  

- Why are there multiple pom.xml files? I thought most projects only need 1.
I think the multi-level pom makes sense is because it might be more convenient to trigger a build command from the top level just like the "python setup.py install" as opposed to making the user switch to one level lower just for the java artefact.  The number of pom.xml files is proportional to the number of modules in a given project.  This allows for building inside the lower level modules only if required ( with some caveats) for quick dev/test cycles.
- What does jep-parent mean? For the artifactId think this is the name of the jar in which case I would prefer just jep. I'm not sure what it means in the name of the iml file though?
The reason I went in for jep-parent is because of the parent and child module relationship. Note that the jep-parent is only a shell and does not produce any "jar" artefact. It exists so that we can control common property definitions at the top level where the main build command would be triggered and all of the child modules inherit those properties. 
The actual artifact that is the jar is named "jep-java-dependencies" as Jep is a union of many constructs and we are clearly calling out that this artefact is for the java components. 
- Are the test classes included in the jars? Ideally the test classes would be omitted from the distributed jars.
Not sure if they are included. My understanding is that they are not included by default and one has to write explicit include constructs to include test classes. I can confirm this after I get access to my development machine later in the day.
-Is the commented out block at the bottom of src/pom.xml significant or was that just for experimentation?
Apologies. An overlook on my part. Updated the pull request accordingly. 

Agree to the rest of the comments:

- Ideally we would be integrating this with Travis once we get the maven central repo approval. 
- Regarding the travis construct to automatically deploy to maven central I am not sure who is the right person in this project who could raise a request for this ? The process required is documented here. http://central.sonatype.org/pages/ossrh-guide.html and I believe one of the committers would be the right person to raise this.
- A travis construct sounds good and is the right approach. I believe we will have to do some scripting to enable a release build once the maven central request gets approved. There are maven plugins which automatically update the release versions if a release is being done for a new version of the artefact. http://maven.apache.org/maven-release/maven-release-plugin. However there are sceptics of these kinds of plugins because of https://dzone.com/articles/why-i-never-use-maven-release

ananthc:
@bsteffensmeier  - The test classes are generally ignored by the maven compiler plugin. However the current source structure is intermixed with actual sources and test sources under a common parent package. I have added specific exclude filter that takes care of this. The test classes are now excluded from the jar that is created on every build.

Techcable:
Personally, I don't like to include the IDE files, since the IDE's automatic changes reates unessicarry churn in the commit history. Hopefully a proper maven setup should allow IntelliJ to infer this info, so including the files would only mask an issue wih the inference.

ndjensen:
Ok, so where we're at:

1. This pull request supports local maven support.  Presumably this is useful to many people/projects?
2. I am not against this pull request since it doesn't mess with setup.py's ease of use for building.
3. We can't seem to get on central maven without owning the domain name, which is not available.  https://issues.sonatype.org/browse/OSSRH-34206  I suggest we defer for now on any attempts to get there.  ([ninia](https://en.wikipedia.org/wiki/Ninia) is a coffeesnake since it's not obvious).
4. I'm somewhat indifferent on the IDE files, it's not clear to me how useful they are to developers using those IDES.  There's no way @bsteffensmeier and I can maintain them so they may become obsolete.
5. I can work on a fork or branch to try and get the source and test folders structured like maven expects, while keeping setup.py building and testing.  I'm not sure when I can get to that.  At that time we'd need help to update the pom.xml but it should be simpler then?
6. I still like the idea of the jar just being named jep-3.7.0.jar.  If someone doesn't build the native library, they just get UnsatisfiedLinkError.  We have no Java dependencies (and intend to keep it that way), it's just a jar and a native library.  I specifically made setup.py build a jar following open source naming conventions.

Presuming a decent number of people will find this local maven support useful, I support it and am thankful @ananthc is taking the time to get it to work.

Techcable:
You could choose a different TLD, since both `ninia.io` and `ninia.us` are available.

ndjensen:
I'm against anything .io due to #29.  Do you know if maven central will let us change the groupId with a major release?  Not that it matters that much.  But the mythical jep4 in my head is going to be awesome.

ndjensen:
Ok, I took a few hours to try and get the source laid out in a maven style while still keeping setup.py and all its commands working as well as before.  Please review:  https://github.com/ndjensen/jep/tree/maven

1. Is the layout right for maven?  I don't know how to test building with maven.  I also didn't make a pom.xml file at this time.
2. If the layout's wrong, please point out what I did wrong.
3. Would it simplify this pull request?
4. Does it make sense?  Jep is weird with crossing the boundary to Python and C, not just Java.

It's on a separate branch on my separate fork, indicating the experimental nature.

ananthc:
@Techcable / @ndjensen / @bsteffensmeier  - Removed the IDE files per advice. 

@ndjensen  - Thanks for proposing the refactor. It is more aligned I guess for future changes. 
The POM.xml that is being proposed will need to be changed. I can  resubmit the pull request after the refactor is merged to master.

A couple of thoughts: 

It will certainly help for this project to be in the maven repo in the immediate future. The reason I state this is because the initial tests for Apache Apex look promising for me. I am yet to complete all the benchmarking code for my tests but I guess it loos very promising. 

Not having this in the central maven repo means that we will have to include the binary jar in the source artefact for the Apache Apex project and this might lead to objections. 

Regarding the 4.x release, I am curious if there is any roadmap or a rough list of items .. I can align some of the feature implementation for Apex accordingly.

ndjensen:
@Techcable I grabbed ninia.us but it turns out you can't get domain privacy on .us domains.  :-(  Now I'm suddenly getting spammed on email and mobile phone and probably address too.  I am trying to delete the domain as fast as I can.   @bsteffensmeier and I were thinking ninia.black.  There is a black coffeesnake.  Thoughts?  The package name would not match the groupId, but the maven central people seem adamant about owning a domain.

@ananthc Jep 4 is somewhat mythical in that I don't know when it will arrive and there is no roadmap.  I want to break backwards compatibility on the Java side (hence 4.0) and refactor the code for cleanliness.  Regarding features, @bsteffensmeier and I are the only active developers of Jep and we only work on it in our spare time.  Right now we are taking a break since the 3.7 release appears to be pretty good.  There will probably be a 3.8 first.

Techcable:
@ndjensen I think `ninia.black` is an **awesome** domain name, and I hope the maven central people finally let you get in! My only issue at this point is how you'd distribute the native library along with the jar, though I guess that's a separate problem.

bsteffensmeier:
We are now set up to use sonatype to push to maven central with the groupId black.ninia. I have pushed the artifacts created from our current build for 3.7 and they are sitting in a staging repository [here](https://oss.sonatype.org/content/repositories/blackninia-1000/).

Can @techcable and/or @ananthc attempt to connect to the staging repository and verify that maven can find the jars and that the work with the native pieces of 3.7? I don't have a project setup with maven so I cannot easily test that everything is working and I think once it is released we would have to change the version to make changes.

If everything looks good in the staging repo then I can release it and it should go to maven central. I think that should be all we need for 3.7. In my opinion we should leave 3.7 alone and try to streamline things on 3.8 by getting the pom checked in and merging it with @ndjensen new directory layout.

ndjensen:
Here's where we're at.

1. @bsteffensmeier set up a 3.7.0 jep on the central maven.
2. I merged the maven relayout of folders into branch dev_3.8.  The current Jep policy is the master branch matches the latest release (in this case 3.7.0) and new changes go on the next dev branch (in this case dev_3.8).  While this is only a build change, I didn't want to introduce risk to 3.7.
3. @bsteffensmeier put an initial pom.xml in dev_3.8.

What else do we need to do?

ananthc:
@ndjensen / @Techcable / @bsteffensmeier  - Apologies for the delayed response on this thread. I have been tied up due to other pressing issues which I could not avoid. 

@bsteffensmeier - I tried traversing through the repository using the browser but could not get through. I tried searching through the search interface of Maven but cannot see it either.  Could you please provide me the artefact details like groupid , artifactid and version so that I can test it by pulling it down into my environment ?

ananthc:
@ndjensen  - I am assuming the next step is to enable a push into the maven central repo by configuring Jenkins/travis to push the built artefact into the appropriate repository ? I am assuming we will be pushing into the 3.8.0-SNAPSHOT repo till we have a gold build for 3.8.0 ?

bsteffensmeier:
@ananthc The staging repo is gone, it looks like sonatype aggressively cleans up the staging repositories. I have created a new staging repository at https://oss.sonatype.org/content/repositories/blackninia-1001/. The groupid is "black.ninia", artefactid is "jep" and version is "3.7.0".

If no one finds a problem with this repo in the next couple of days I intend to push it to maven central.

ananthc:
@bsteffensmeier - Thanks for the info. I have verified the artefact. I have created a local maven project, added the staging repo that you gave above to be searched for and put a dependency on jet 3.7.0 in the local project. The project built successfully. Thanks for the info.

bsteffensmeier:
I just promoted the staging repo into maven central so it should now be possible to use jep 3.7.0 from maven central.

ndjensen:
Should we close and/or cancel this pull request?  Is there anything left to do?

ananthc:
@ndjensen  - I believe this PR can be closed. I am using the maven repo version artefact for 3.7 and I believe the code refactor for 3.8.0 has the required pom already.

PR #95, opened 2017-09-25
jep library loading from Classpath

deicon:
In order to bundle the jep.dll/.so within a java app,
the lib will be searched within the root of the classpath as either
jep.dll or jep.so.

See static initializer of Jep.java

This Patch includes Work taken from
https://www.adamheinrich.com/blog/2012/12/how-to-load-native-jni-library-from-jar/is 
This way the jep.so can be bundled with the Java App. jep installation still required though. 

It would be ideal, to just depend on a python installation without the need of extra installation steps.

Not sure if this is of any use to anyone else than me, but wanted to share anyways.

### DIFF
--- a/src/jep/Jep.java
+++ b/src/jep/Jep.java
@@ -1,26 +1,26 @@
 /**
  * Copyright (c) 2017 JEP AUTHORS.
- *
+ * <p>
  * This file is licensed under the the zlib/libpng License.
- *
+ * <p>
  * This software is provided 'as-is', without any express or implied
  * warranty. In no event will the authors be held liable for any
  * damages arising from the use of this software.
- * 
+ * <p>
  * Permission is granted to anyone to use this software for any
  * purpose, including commercial applications, and to alter it and
  * redistribute it freely, subject to the following restrictions:
- * 
- *     1. The origin of this software must not be misrepresented; you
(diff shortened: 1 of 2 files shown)


ndjensen:
The link doesn't work.

We can't package the native library with the jar since we don't know what version of Python it was built against and whether or not it was compiled with numpy support.  It's possible we could make the build smart enough to put the native library in the jar and update the code as you have suggested, but the code would need to be more cross platform.  We'd also want to go a step further and try to get the Python code inside the jar too.  I presume you're putting the built library in the jar manually?

deicon:
Actually I have integrated it using a mvn build which bundles the jar with the library. I might as well spice this pom.xml up and check it in as well. 

I agree that all this only makes any sense if it also contains all needed python files as well. I am no expert with python at all. So I have no idea if it's possible to bundle them as well.

ndjensen:
dev_3.8 branch has better maven support.  You should look at it instead of master.  Master matches the currently released code (3.7.0).

bsteffensmeier:
This type of functionality would require far more work before it should be merged; such as including required python files and including support for various python versions and/or supporting python binary compatibility and the build changes to support all of that. With the release of 3.8 this request is now conflicting and inactive so I am closing it.

PR #102, opened 2017-12-06
Make ClassList::loadClassPath inspect manifest classpath

djbarker:
`ClassList::loadClassPath` currently ignores all non `.class` files inside `.jar` files. This means that any dependencies in the manifest do not get loaded which can lead to import errors. This PR fixes that.

### DIFF
--- a/src/jep/ClassList.java
+++ b/src/jep/ClassList.java
@@ -104,7 +104,7 @@ public class ClassList implements ClassEnquirer {
                         String[] relativePaths = classpath.split(" ");
 
                         for(String relativePath : relativePaths) {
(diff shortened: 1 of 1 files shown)


bsteffensmeier:
1. Can you move or resubmit this in the dev_3.8 branch, see the [Contributing Guidelines.](https://github.com/ninia/jep/blob/master/.github/CONTRIBUTING.md)
2. Is there any concern that jars will contain duplicate entries that we will process more than once. Along the same lines do we need to be paranoid about checking for circular references that would infinite loop us?

djbarker:
1. Sorry, I'll redo this from dev_3.8 which should fix the conflicts. 
2. Very valid point, I'll investigate.

PR #103, opened 2017-12-08
Make ClassList::loadClassPath inspect manifest classpath

djbarker:
`ClassList::loadClassPath` currently ignores all non `.class` files inside `.jar` files. This means that any dependencies in the manifest do not get loaded which can lead to import errors. This PR fixes that.

### DIFF
--- a/src/main/java/jep/ClassList.java
+++ b/src/main/java/jep/ClassList.java
@@ -78,12 +78,13 @@ public class ClassList implements ClassEnquirer {
         Set<String> seen = new HashSet<>();
 
         while(tok.hasMoreTokens()) {
-            queue.add(tok.nextToken());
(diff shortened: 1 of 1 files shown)


djbarker:
Hi Ben, will you merge this in or do I need to do anything? 

On a related note; when will Jep 3.8.0 be released?

bsteffensmeier:
We don't have a specific release roadmap but as a general rule we target about one release a year, so all I can really say is that 3.8.0 will be released at some point in 2018. Since 3.7.0 came out in August I wasn't really planning on pushing for it until June. In addition to the arbitrary time frames we also try to capture a reasonable number of changes in a release so contributions like this can help inspire a release sooner.

PR #185, opened 2019-04-29
Always return a `PyObject` if requested type is `PyObject`.

jeremyrsmith:
Adds a higher-priority check if the requested type is `PyObject`. If it is, a `PyObject` will be returned.

Fixes an issue with numpy arrays – currently, they cannot be retrieved as a `PyObject`, and attempting to do so results in a `TypeError` (since numpy conversion logic has no escape hatch if expected type is not a Java array nor Java `NDArray`).

I did this because I think it should always be possible to ask for (and get) a `PyObject`. For example, I may want to operate on large numpy arrays (or other large data structures) that were created from Python without copying to the JVM.

### DIFF
--- a/src/main/c/Jep/convert_p2j.c
+++ b/src/main/c/Jep/convert_p2j.c
@@ -865,6 +865,16 @@ jobject PyObject_As_jobject(JNIEnv *env, PyObject *pyobject,
         if ((*env)->IsAssignableFrom(env, pyjobject->clazz, expectedType)) {
             return (*env)->NewLocalRef(env, pyjobject->object);
         }
+    } else if (PyCallable_Check(pyobject)) {
+        if (isFunctionalInterfaceType(env, expectedType)) {
+            return PyCallable_as_functional_interface(env, pyobject, expectedType);
+        } else if (PyErr_Occurred()) {
+            return NULL;
+        } else if ((*env)->IsAssignableFrom(env, JPYCALLABLE_TYPE, expectedType)) {
+            return PyCallable_As_JPyCallable(env, pyobject);
+        }
+    } else if ((*env)->IsAssignableFrom(env, expectedType, JPYOBJECT_TYPE)) {
+        return PyObject_As_JPyObject(env, pyobject);
 #if PY_MAJOR_VERSION < 3
     } else if (PyString_Check(pyobject)) {
         return pystring_as_jobject(env, pyobject, expectedType);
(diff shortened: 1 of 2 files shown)


src/main/c/Jep/convert_p2j.c
```diff
@@ -865,6 +865,16 @@ jobject PyObject_As_jobject(JNIEnv *env, PyObject *pyobject,
         if ((*env)->IsAssignableFrom(env, pyjobject->clazz, expectedType)) {
             return (*env)->NewLocalRef(env, pyjobject->object);
         }
+    } else if (PyCallable_Check(pyobject)) {
```
jeremyrsmith: Moved up the `PyCallable` check, since it needs to happen before `PyObject` test (since `PyCallable` extends `PyObject`).

src/main/c/Jep/convert_p2j.c
```diff
@@ -865,6 +865,16 @@ jobject PyObject_As_jobject(JNIEnv *env, PyObject *pyobject,
...
+        if (isFunctionalInterfaceType(env, expectedType)) {
+            return PyCallable_as_functional_interface(env, pyobject, expectedType);
+        } else if (PyErr_Occurred()) {
+            return NULL;
+        } else if ((*env)->IsAssignableFrom(env, JPYCALLABLE_TYPE, expectedType)) {
+            return PyCallable_As_JPyCallable(env, pyobject);
+        }
+    } else if ((*env)->IsAssignableFrom(env, expectedType, JPYOBJECT_TYPE)) {
```
jeremyrsmith: This check is the opposite of the other – it's checking whether the *expected type* is `PyObject`, not whether we can *assign* a `PyObject` to the expected type (which is typically a fall-through case when expected type is `Object`).

ndjensen:
Thank you for submitting this on the dev_3.9 branch.  I agree with your reasoning.

src/main/c/Jep/convert_p2j.c
bsteffensmeier: I think `IsSameObject` would be better since it doesn't work correctly on subclasses of PyObject.

jeremyrsmith: Good call – changed to `IsSameObject` (this allow means that the `PyCallable` check can be moved back down to its original position, since it will no longer trigger the `PyObject` branch)

jeremyrsmith:
Any ideas of when this might make it into a release?

ndjensen:
@bsteffensmeier and I were planning on releasing 3.9 this summer but haven't had time to work on it. We'll get a slim 3.9 released without much new content or cherry pick these changes into 3.8 within the next few weeks.

jeremyrsmith:
Awesome, thank you! If there's other work you wanted to do for 3.9 I'd be happy to take a crack at a couple of issues if you folks don't have time. We're heavily relying on jep for python support in our open source project so we'd be happy to contribute!

PR #186, opened 2019-04-29
Add `callAs` overloads to `PyCallable`

jeremyrsmith:
Adds overloads to `PyCallable` which mirror the respective `call` methods, but allow passing an expected `Class` for the result.

### DIFF
--- a/src/main/java/jep/python/PyCallable.java
+++ b/src/main/java/jep/python/PyCallable.java
@@ -77,8 +77,23 @@ public class PyCallable extends PyObject {
      *             if an error occurs
      */
     public Object call(Object... args) throws JepException {
+        return callAs(Object.class, args);
+    }
+
+    /**
+     * Invokes this callable with the args in order, converting the return value to the given class.
+     *
+     * @param expectedType
+     *            The expected return type of the invocation
+     * @param args
+     *            args to pass to the function in order
+     * @return an value of the given type
+     * @throws JepException
+     *             if an error occurs
+     */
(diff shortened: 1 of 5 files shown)


jeremyrsmith:
Oops, rebased onto dev branch

ndjensen:
This is great!  I had no idea if anyone was actually using PyCallable or not.  I wrote it as a cool idea but didn't know if it was actually used.  I like your idea of allowing the caller to specify the return type.  My only concern is the usage of the word "as" is somewhat confusing.  That said, I haven't come up with a better name for the methods.

src/main/java/jep/python/PyCallable.java
```diff
@@ -91,8 +106,23 @@ public Object call(Object... args) throws JepException {
...
+    }
+
+    /**
+     * Invokes this callable with keyword args.
+     * 
+     * @param expectedType
+     *            The expected return type of the invocation
+     * @param kwargs
+     *            a Map of keyword args
+     * @return an {@link Object} value
```
bsteffensmeier: You are inconsistent on the @return javadoc for the callAs. They should all be "a value of the given type"

jeremyrsmith:
@ndjensen I'm not a huge fan of `callAs` either, but since some `call` methods take varargs of `Object` it seemed prudent to name these something else rather than having somewhat ambiguous overloads.

src/main/java/jep/python/PyCallable.java
jeremyrsmith: :+1: fixed

Issue #210, opened 2019-10-29
JNI error immediately after running "jep" on macOS


gosha1128:
**Describe the bug**

I installed jep successfully using Anaconda3 pip.  When i run "jep" I get a JNI error.

Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: jep/Run has been compiled by a more recent version of the Java Runtime (class file version 56.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:495)

**To Reproduce**
Install via pip.  Then run jep.

**Expected behavior**
I expected to see a CLI prompt.

**Environment (please complete the following information):**
- OS Platform, Distribution, and Version:  MacOS 10.12.6
- Python Distribution and Version:  Anaconda Python3
- Java Distribution and Version: /Library/Java/JavaVirtualMachines/jdk1.8.0_221.jdk/Contents/Home/
- Jep Version:  
jep in /Users/george/anaconda3/lib/python3.7/site-packages (3.9.0)
- Python packages used (e.g. numpy, pandas, tensorflow):
didn't get that far

**Additional context**
I'm probably being a bonehead.  Apologies in advance.

bsteffensmeier:
[...]

PR #212, opened 2019-11-06
build: decode strings before split

tschoonj:
This fixes the build on macOS with python 3 if MACOSX_DEPLOYMENT_TARGET is set.
I can confirm that it still works with python 2.7 after applying this patch.


ndjensen:
Thank you for contributing!  Can you please make the pull request to the dev_3.9 branch?  We like to keep the master branch as the stable/release branch and do development on dev branches.  Your fix will probably go in a 3.9.1 release.

Also, since you are familiar with building Jep on macOS, if you have any knowledge to share on #210, that would also be appreciated.

PR #234, opened 2020-02-26
Postponed Jep initialization to support decouple init/execution of interpreter on different threads

aitorhh:
Hi,

I have added small modifications to be able to support a postponed initialization of the Jep Instance. 
There is a need to have a postponed initialization when when the create of the Jep, and its usage are done on different threads. 

The quick solution I provided is by exposing a boolean into the Jep constructor. Howeve, I see that the latest versions has the *useSubInterpreter*, which then means that this version will not be backwards compatibly.

BR,

//Aitor

### DIFF
--- a/src/main/java/jep/Jep.java
+++ b/src/main/java/jep/Jep.java
@@ -6,19 +6,19 @@
  * This software is provided 'as-is', without any express or implied
  * warranty. In no event will the authors be held liable for any
  * damages arising from the use of this software.
- * 
+ *
  * Permission is granted to anyone to use this software for any
  * purpose, including commercial applications, and to alter it and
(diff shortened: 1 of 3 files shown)


bsteffensmeier:
I don't understand why it would be necessary to construct an Interpreter on a different thread than it is being used on. To avoid confusion I don't think Jep should provide any API that encourages developers to pass a Interpreter between threads. If lazy initialization is needed it should be possible to get the same behaviour with a factory-like object outside of Jep that can hold a JepConfig and construct the Jep instance on the appropriate thread.

aitorhh:
Agree! :) Closing the pull request.

Issue #202, opened 2019-08-27
Features for Jep 4.0


ndjensen:
Jep 4.0 is planned for release in 2020 **or** later.  4.0 aims to clean out old API and add more notable API changes.  The ideas listed here are not a commitment, just a tracking point.  Some ideas are:

- Remove deprecated methods.  Make Jep type abstract so subclass must be instantiated.
- Remove Python 2 support. Will clean up C code.
- Remove Python 3.3, 3.4, 3.5 support. Will clean up C code.
- Change fallback last resort behavior of Jep.getValue() returning a string of the pyobject to instead return a PyObject object.
- JepException extends RuntimeException instead of Exception. Checked exceptions are not great.
- PyJType: The idea that the type of a PyJObject is a true Python Type, i.e. a Python type maps to a Java class.  For example, type(ArrayList) would be `java.util.ArrayList` instead of `PyJObject`.  Rather complicated, major roadblock is that `if (type(Object()) == Object().getClass()` works then `isinstance(Object.getClass(), Class)` won't work, and vice versa.  What to do?
- Better support for JPyObjects aka Java jep.python.PyObjects such as JPyList (jep.python.PyList extends jep.python.PyObject), JPyDict (jep.python.PyDict extends jep.python.PyObject), etc.  Also known as the reverse of the PyJList, PyJMap, PyJIterator, etc
- PEJ - Python Embedded Java: The reverse of Jep, a Python process with a JVM inside it.  Other open source projects already do this.
- Pluggable Java <-> Python type converters.  We have builtin converters for PyList <-> Java ArrayList, PyDict <-> Java HashMap, Python primitives <-> Java primitives.  We have custom converters for numpy support.  Converters should be detectable at build time and included as desired.  For example, someone could write a `pandas DataFrame` converter to a Java object, and contribute it back to the community.  But users would get to choose whether to build with pandas support or not.  Another example from a mailing list post is to convert a Python `numpy ndarray` to a `List<Float>` instead of a Java `NDArray`.  Users could write custom converters and choose which converters to build with.
- Simplify build by removing support for Python 2, Java 1.7, and Apple's JVM.  Contemplate moving to setuptools build, but we have trouble getting multiple environments to test in and could use help if we alter the build.
- Better API for pyjarray and pinning arrays (for example `with pyjarray.pin(): for loop`).  Pinning is slowing down primitive pyjarrays unless you're looping over contents.
[...]

PR #255, opened 2020-06-03
Removed python2 support

JCWasmx86:
This pull request removes all #if PY_MAJOR_VERSION checks, so Python3+ is the only supported version. Furthermore this leads to cleaner code. Removing the support for Python 2 was proposed in #202

### DIFF
--- a/src/main/c/Include/jep_platform.h
+++ b/src/main/c/Include/jep_platform.h
@@ -74,34 +74,23 @@
     */
     #define JLOCAL_REFS 16
 
-    /* Python 2 compatibility */
-    #if PY_MAJOR_VERSION < 3
-        #define Py_hash_t long
-    #endif
+    // see https://mail.python.org/pipermail/python-porting/2012-April/000289.html
+    #define Py_TPFLAGS_HAVE_ITER 0
 
-    /* Python 3 compatibility */
-    #if PY_MAJOR_VERSION >= 3
+    /* Python 3 does not support integers, only longs */
+    #define PyInt_AsLong(i)                   PyLong_AsLongLong(i)
+    #define PyInt_AS_LONG(i)                  PyLong_AsLongLong(i)
+    #define PyInt_Check(i)                    PyLong_Check(i)
+    #define PyInt_FromLong(i)                 PyLong_FromLongLong(i)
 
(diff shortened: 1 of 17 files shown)


src/main/c/Include/jep_platform.h
```diff
@@ -74,34 +74,23 @@
...
-        #define PyString_Size(str)                PyUnicode_GetLength(str)
-        #define PyString_GET_SIZE(str)            PyUnicode_GET_LENGTH(str)
-
-    #endif // Python 3 compatibility
-
+    // see https://mail.python.org/pipermail/python-porting/2012-April/000289.html
+    #define Py_TPFLAGS_HAVE_ITER 0
+
+    /* Python 3 does not support integers, only longs */
+    #define PyInt_AsLong(i)                   PyLong_AsLongLong(i)
```
ndjensen, line 81: Redefining these methods for Python 3 was all to support being compatible with Python 2 and Python 3.  Since Python 2 is being removed, we should no longer have any code calling `PyInt_AsLong(i)`, but we might.  Any remaining calls to `PyInt_AsLong(i)` in Jep src should be updated to be PyLong_AsLongLong(i) to match the same behavior.  This kind of update can be done for all these method replacements.  This should be done for maximum cleanliness and clarity of code.

ndjensen:
Oh we need to remove Python 2.7 from the AppVeyor and Travis CI files to get the build to pass.

src/main/c/Include/jep_platform.h
```diff
@@ -74,34 +74,23 @@
...
-
-        #define PyString_AsString(str)            PyUnicode_AsUTF8(str)
-        #define PyString_AS_STRING(str)           PyUnicode_AsUTF8(str)
-        #define PyString_Size(str)                PyUnicode_GetLength(str)
-        #define PyString_GET_SIZE(str)            PyUnicode_GET_LENGTH(str)
-
-    #endif // Python 3 compatibility
-
+    // see https://mail.python.org/pipermail/python-porting/2012-April/000289.html
+    #define Py_TPFLAGS_HAVE_ITER 0
```
bsteffensmeier, line 78: Py_TPFLAGS_HAVE_ITER is also just for python 2 compatibility, since we are removing compatibility it would be better to just remove the reference from pyjarray.c

PR #256, opened 2020-06-04
Removing Python 2 

JCWasmx86:
In this pull request I removed all checks for the major python version, with only support for Python 3.
Furthermore I removed all Macros used for compatibility.


ndjensen:
Thank you!

PR #257, opened 2020-06-05
More checks for Python 2 are removed

JCWasmx86:
I removed some checks in the tests that test for Python 2.
Furthermore Python 3.5 is now the oldest supported version.


src/main/java/jep/Jep.java
```diff
@@ -605,384 +396,41 @@ private native void set(long tstate, String name, Object v)
...
-     *             if an error occurs
-     */
-    @Deprecated
-    public void set(String name, boolean v) throws JepException {
-        isValidThread();
-
-        set(tstate, name, v);
-    }
-
-    private native void set(long tstate, String name, boolean v)
```
bsteffensmeier, line 651: When you delete native methods you should also delete the implementation in jep.c

src/main/java/jep/Jep.java
```diff
@@ -497,26 +353,6 @@ public Object getValue(String str) throws JepException {
...
-     *             if an error occurs
-     */
-    @Deprecated
-    public byte[] getValue_bytearray(String str) throws JepException {
-        isValidThread();
-
-        return getValue_bytearray(this.tstate, str);
-    }
-
     private native byte[] getValue_bytearray(long tstate, String str)
```
bsteffensmeier, line 356: Should remove the private methods when they aren't used anymore

README.rst
```diff
@@ -44,7 +44,7 @@ Building and installing require the JDK, Python, and optionally numpy to be inst
 
 Dependencies
 ------------
-* Python 2.7, 3.3, 3.4, 3.5, 3.6, 3.7, or 3.8
+* Python 3.5, 3.6, 3.7, or 3.8
 * Java >= 1.7
```
bsteffensmeier, line 48: Java needs to change to 1.8 here

src/main/java/jep/Jep.java
JCWasmx86: Fixed

src/main/java/jep/Jep.java
JCWasmx86: That was removed by accident, I added it again.

src/main/java/jep/Jep.java
```diff
@@ -497,103 +353,30 @@ public Object getValue(String str) throws JepException {
...
-
     @Override
     public void set(String name, Object v) throws JepException {
         isValidThread();
 
         if (v instanceof Class) {
             set(tstate, name, (Class<?>) v);
         } else if (v instanceof String) {
-            set(name, ((String) v));
+            set(tstate,name, ((String) v));
```
bsteffensmeier, line 365: I think we should just pass the Object to the native code and remove this if block. We used to have similar checks/conversions in many different places and the inconsistencies were a problem so I have been consolidating the checks ad this one should be the next to go.

The conversion code already has the same type of conversion in c so this is just redundant and it will simplify things to remove many of the native methods from this class.

src/main/c/Jep/jep.c
```diff
@@ -61,7 +61,7 @@ JNI_OnUnload(JavaVM *vm, void *reserved)
 /*
  * Class:     jep_Jep
  * Method:    init
- * Signature: (Ljava/lang/ClassLoader;Z)I
+ * Signature: (Ljava/lang/ClassLoader;Z)J
```
ndjensen, line 64: Wow, good catch on these method signature comments.

src/main/java/jep/Jep.java
ndjensen: Are you saying to remove all these ifs and just have the code in the final else, i.e. `set(tstate, name, object)`?  No more instanceof and casting?  That would be most clean in my opinion.

src/main/java/jep/Jep.java
```diff
@@ -86,52 +86,13 @@
      * you make use of CPython extensions (e.g. numpy) that use the GIL, then
      * this gets even more risky and can potentially deadlock.
      */
-    private final static ThreadLocal<Boolean> threadUsed = new ThreadLocal<Boolean>() {
+    private static final ThreadLocal<Boolean> threadUsed = new ThreadLocal<Boolean>() {
```
ndjensen, line 89: Can we use the diamond operator on the right side, i.e. `new ThreadLocal<>()`?

JCWasmx86: The diamond operator would work here.

src/main/java/jep/Jep.java
bsteffensmeier: Yes, that is what is happening in the later patchsets and it is still passing tests on AppVeyor which is what I expected.

src/main/c/Jep/pyembed.c
```diff
@@ -212,8 +212,7 @@ void pyembed_preinit(JNIEnv *env,
         wchar_t* homeForPython = malloc((length + 1) * sizeof(wchar_t));
         mbstowcs(homeForPython, homeAsUTF, length + 1);
 #endif
-        (*env)->ReleaseStringUTFChars(env, pythonHome, homeAsUTF);
-
+		(*env)->ReleaseStringUTFChars(env, pythonHome, homeAsUTF);
```
ndjensen, line 215: Oops, new indent was inserted here.

ndjensen:
Since you removed most of the native calls of `Jep.set(String, specificType)` in favor of making everything call `Jep.set(String, Object)`, you were able to remove native methods in jep.c.  Those in turn called into pyembed.c and called `pyembed_setparameter_specificType()`.  I think most of those `pyembed_setparameter_specificType` are now no longer used and dead code.

ndjensen:
Thank you!

JCWasmx86:
You helped me a lot too, thank you

PR #264, opened 2020-06-11
Cleaned up the code, fixed memory leaks.

JCWasmx86:
This are the changes from #257 for the branch dev_4.0. (The new commits start from 
#379df54)
Furthermore, I fixed some small memory leaks.
I removed all sizeof(char) expressions, as it is guaranteed from the C specification, that it is 1.

### DIFF
--- a/src/main/c/Jep/jep_exceptions.c
+++ b/src/main/c/Jep/jep_exceptions.c
@@ -258,7 +258,7 @@ int process_py_exception(JNIEnv *env)
 
                         // remove the .py to look more like a Java StackTraceElement
                         namelen = (int) strlen(charPyFile);
-                        charPyFileNoExt = malloc(namelen + 1);
+                        charPyFileNoExt = malloc(sizeof(char) * (namelen + 1));
                         strcpy(charPyFileNoExt, charPyFile);
                         lastDot = strrchr(charPyFileNoExt, '.');
                         if (lastDot != NULL) {
--- a/src/main/c/Jep/jep_exceptions.c
+++ b/src/main/c/Jep/jep_exceptions.c
@@ -258,7 +258,7 @@ int process_py_exception(JNIEnv *env)
 
                         // remove the .py to look more like a Java StackTraceElement
                         namelen = (int) strlen(charPyFile);
-                        charPyFileNoExt = malloc(sizeof(char) * (namelen + 1));
+                        charPyFileNoExt = malloc(namelen + 1);
                         strcpy(charPyFileNoExt, charPyFile);
                         lastDot = strrchr(charPyFileNoExt, '.');
                         if (lastDot != NULL) {
@@ -266,7 +266,7 @@ int process_py_exception(JNIEnv *env)
                         }
 
                         // remove the dir path to look more like a Java StackTraceElement
-                        charPyFileNoDir = malloc(sizeof(char) * (namelen + 1));
+                        charPyFileNoDir = malloc(namelen + 1);
                         lastBackslash = strrchr(charPyFile, FILE_SEP);
                         if (lastBackslash != NULL) {
                             strcpy(charPyFileNoDir, lastBackslash + 1);
--- a/src/main/c/Jep/jep_numpy.c
+++ b/src/main/c/Jep/jep_numpy.c
@@ -589,8 +589,6 @@ PyObject* convert_jdndarray_pyndarray(JNIEnv *env, PyObject* pyobj)
     data = (*env)->CallObjectMethod(env, obj, dndarrayGetData);
     if (process_java_exception(env) || !data) {
         free(dims);
-        //Is this necessary?
-        (*env)->DeleteLocalRef(env, data);
         return NULL;
     }
 
--- a/src/main/c/Jep/jep_numpy.c
+++ b/src/main/c/Jep/jep_numpy.c
@@ -393,7 +393,6 @@ static jobject convert_pyndarray_jndarray(JNIEnv *env, PyObject *pyobj)
     for (i = 0; i < ndims; i++) {
         jdims[i] = (jint) dims[i];
     }
-
     jdimObj = (*env)->NewIntArray(env, ndims);
     if (process_java_exception(env) || !jdimObj) {
         free(jdims);
@@ -589,6 +588,9 @@ PyObject* convert_jdndarray_pyndarray(JNIEnv *env, PyObject* pyobj)
     // get the primitive array and convert it
     data = (*env)->CallObjectMethod(env, obj, dndarrayGetData);
     if (process_java_exception(env) || !data) {
+        free(dims);
+        //Is this necessary?
+        (*env)->DeleteLocalRef(env, data);
         return NULL;
     }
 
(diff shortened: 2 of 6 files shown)


JCWasmx86:
Does anyone have an idea, why the build is failing for the first two checks?

src/main/c/Objects/pyjmethod.c
```diff
@@ -179,7 +179,6 @@ int PyJMethod_CheckArguments(PyJMethodObject* method, JNIEnv *env,
                                 method->parameters, parampos);
 
         if (process_java_exception(env) || !paramType) {
-            match = 0;
```
bsteffensmeier, line 182: I think we should set matchTotal to -1 and document in the header that negative return values indicate errors.

src/main/c/Objects/pyjmethod.c
```diff
@@ -188,7 +187,6 @@ int PyJMethod_CheckArguments(PyJMethodObject* method, JNIEnv *env,
         match = pyarg_matches_jtype(env, param, paramType, paramTypeId);
         (*env)->DeleteLocalRef(env, paramType);
         if (PyErr_Occurred()) {
-            match = 0;
```
bsteffensmeier, line 191: I think this should also set matchTotal to -1.

ndjensen:
@JCWasmx86, are you going to resumbit some of these changes in another pull request?  Please let us know one way or the other.  If you want to discuss changes over email, you can send an email to my github username at gmail, I have the same username on github and gmail.

JCWasmx86:
Sorry, I wasn't able to rebase my changes, (I'm quite new to github).

bsteffensmeier:
It should be possible with rebase command as described [here](https://makandracards.com/makandra/45964-git-rebase-dependent-feature-branch-after-squash). Or if you checkout the current dev_4.0 you could cherry-pick your new changes onto that branch with `git cherry-pick 379df54^..0eab6b5`

JCWasmx86:
Ok, I fixed all issues (At least I hope so) and applied all changes again.

src/main/c/Jep/jep_exceptions.c
```diff
@@ -258,15 +258,15 @@ int process_py_exception(JNIEnv *env)
 
                         // remove the .py to look more like a Java StackTraceElement
                         namelen = (int) strlen(charPyFile);
-                        charPyFileNoExt = malloc(sizeof(char) * (namelen + 1));
+                        charPyFileNoExt = malloc(namelen + 1);
```
ndjensen, line 261: I like the two sizeof(char) in this here because I had to get sizeof(type) around nearly everything to support more platforms and 32-bit vs 64-bit.  I agree it doesn't make a difference, I just like the style of it.

src/main/c/Jep/jep_numpy.c
```diff
@@ -589,6 +588,9 @@ PyObject* convert_jdndarray_pyndarray(JNIEnv *env, PyObject* pyobj)
     // get the primitive array and convert it
     data = (*env)->CallObjectMethod(env, obj, dndarrayGetData);
     if (process_java_exception(env) || !data) {
+        free(dims);
+        //Is this necessary?
+        (*env)->DeleteLocalRef(env, data);
```
ndjensen, line 593: Good catch on free(dims).  I'm not sure what happens if you `(*env)->DeleteLocalRef(env, NULL);`.  It could be bad?  If we don't know we shouldn't add that line (and the comment).

src/main/c/Jep/pyembed.c
```diff
@@ -450,6 +449,10 @@ void pyembed_startup(JNIEnv *env, jobjectArray sharedModulesArgv)
                 PyEval_ReleaseThread(mainThreadState);
                 (*env)->PopLocalFrame(env, NULL);
                 THROW_JEP(env, "Received null argv.");
+                for(j = 0; j < i; j++) {
```
ndjensen, line 452: Good catch.  Side note, I hate the letter `j` because it looks so close to `i` at times.  That caught me once and I swore never again so I use `k`.

src/main/c/Jep/pyembed.c
```diff
@@ -613,6 +616,7 @@ intptr_t pyembed_thread_init(JNIEnv *env, jobject cl, jobject caller,
         if (mod_main == NULL) {
             THROW_JEP(env, "Couldn't add module __main__.");
             PyEval_ReleaseThread(jepThread->tstate);
+            free(jepThread);
```
ndjensen, line 619: Good catch.

src/main/c/Jep/pyembed.c
```diff
@@ -505,7 +508,7 @@ int pyembed_is_version_unsafe(void)
         char *msg;
         JNIEnv *env = pyembed_get_env();
 
-        msg = malloc(sizeof(char) * 200);
+        msg = malloc(200);
```
ndjensen, line 511: I still like sizeOf.

src/main/c/Objects/pyjbuffer.c
```diff
@@ -47,7 +47,7 @@ static const struct bufferdescr descriptors[] = {
     {&JINTBUFFER_TYPE, &java_nio_IntBuffer_order, sizeof(jint), sizeof(int), "i", "<i", ">i"},
     {&JLONGBUFFER_TYPE, &java_nio_LongBuffer_order, sizeof(jlong), sizeof(long long), "q", "<q", ">q"},
     {&JSHORTBUFFER_TYPE, &java_nio_ShortBuffer_order, sizeof(jshort), sizeof(short), "h", "<h", ">h"},
-    {&JCHARBUFFER_TYPE, &java_nio_CharBuffer_order, sizeof(jchar), sizeof(char), "H", "<H", ">H"},
+    {&JCHARBUFFER_TYPE, &java_nio_CharBuffer_order, sizeof(jchar), 1, "H", "<H", ">H"},
```
ndjensen, line 50: Ah, me and sizeof.

src/main/c/Objects/pyjmonitor.c
```diff
@@ -64,7 +64,7 @@ int PyJMonitor_Check(PyObject *obj)
 static PyObject* pyjmonitor_enter(PyObject* self, PyObject* args)
 {
     PyJMonitorObject *monitor = (PyJMonitorObject*) self;
-    JNIEnv           *env     = env = pyembed_get_env();
+    JNIEnv           *env     = pyembed_get_env();
```
ndjensen, line 67: Good catch.

src/main/c/Objects/pyjmonitor.c
```diff
@@ -94,7 +94,7 @@ static PyObject* pyjmonitor_enter(PyObject* self, PyObject* args)
 static PyObject* pyjmonitor_exit(PyObject* self, PyObject* args)
 {
     PyJMonitorObject *monitor  = (PyJMonitorObject*) self;
-    JNIEnv           *env      = env = pyembed_get_env();
+    JNIEnv           *env      = pyembed_get_env();
```
ndjensen, line 97: Good catch.

src/main/c/Jep/jep_numpy.c
JCWasmx86: I tested it, if I pass NULL to DeleteLocalRef, nothing happens (at least on Java 14). I will remove this unnecessary call.

JCWasmx86:
I used two tools:
scan-build (From the clang-tools package)
gcc's option -fanalyzer (https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html), GCC 10
There are around 24 warnings from these tools. Maybe we could remove these warnings and then add it to appveyor, in order to have an even better code quality?

PR #265, opened 2020-06-16
Don't use PyUnicode_GET_SIZE in Python 3.

methane:
It is a deprecated API.  Use PyUnicode_GET_LENGTH instead.


bsteffensmeier:
We only update master when we do a release, please submit this change to the dev_4.0 branch.

methane:
OK, I sent #266 instead.

Issue #267, opened 2020-06-18
Warning on platform depencency


Reissner:
**Describe the problem**
command mvn install yields 
```
[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
```
which is easily to resolve: 
just add definition of encoding in pom. 
My environment is linux but i think this is a quire general problem. 

Sent already a pull request.

valoeghese:
should this issue be closed?

PR #268, opened 2020-06-18
issue 267: added specification of encoding to avoid platform dep.

Reissner:
Resolve issue #267: added just one line in pom


Reissner:
#268

PR #275, opened 2020-07-29
Attempt to load pythonXX.dll from within PYTHONHOME

pacher:
Using Anaconda on Windows, jep correctly loads `jep.dll` if PYTHONHOME is set to the root of the desired conda environment.
But then it fails with `UnsatisfiedLinkError: jep.dll: Can't find dependent libraries.` if PYTHONHOME is not in java.library.path

This does not happen when java app is started from conda command prompt since conda adds env root to `Path` environment variable. But using command prompt is not always convenient. For example I could not find a way to run java app in conda prompt from within Intellij IDE.

For Mac and Linux jep tries to mitigate similar issue by attempting to load libpython from within PYTHONHOME. This PR adds the same attempt to load `pythonXX.dll` for Windows.

Open questions:
- Jep works if `python38.dll` is loaded in my case. There is also `python3.dll` there which does not work. Therefore proposed solution is looking for `pythonXX.dll` where XX is two digits minimum. What exactly corresponds to `libpython` on windows? Because `python38.dll` was found just by trial and error.
- It was not tested for python2 and who knows how it will look in python4. So it might make sense to search for python3X.dll instead of pythonXX.dll since IIRC jep is going to drop support for python2 in future releases.
- There is an option to add base environment to `Path` during Anaconda installation. So it is possible that the user will set PYTHONHOME to the desired environment, but jep will load "wrong" python dll from the base environment. Maybe these attempts to load libpython from within PYTHONHOME should be performed by default before loading jep? And not only when jep fails with UnsatisfiedLinkError?

### DIFF
--- a/src/main/java/jep/LibraryLocator.java
+++ b/src/main/java/jep/LibraryLocator.java
@@ -213,14 +213,16 @@ final class LibraryLocator {
                         System.load(libraryFile.getAbsolutePath());
                     } catch (UnsatisfiedLinkError e) {
                         /*
-                         * This is almost always caused because libpython isn't
+                         * This is almost always caused because libpython or pythonXX.dll isn't
                          * found, so try to figure out the exact libpython that
-                         * is needed and look in PYTHONHOME.
+                         * is needed and look in PYTHONHOME. Otherwise look in PYTHONHOME for pythonXX.dll
                          * 
                          */
                         Matcher m = Pattern.compile("libpython[\\w\\.]*")
                                 .matcher(e.getMessage());
                         if (m.find() && findPythonLibrary(m.group(0))) {
+                            System.load(libraryFile.getAbsolutePath());
+                        } else if (findPythonLibraryWindows()) {
                             System.load(libraryFile.getAbsolutePath());
                         } else {
                             throw e;
@@ -254,6 +256,31 @@ final class LibraryLocator {
                 File libraryFile = new File(libDir, libraryName);
                 if (libraryFile.exists()) {
                     System.load(libraryFile.getAbsolutePath());
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Attempt to load pythonXX.dll from within PYTHONHOME
+     *
+     * @return true if pythonXX.dll was found and loaded.
+     */
+    private boolean findPythonLibraryWindows() {
+        String pythonHome = this.pythonHome;
+        if (pythonHome == null && !ignoreEnv) {
+            pythonHome = System.getenv("PYTHONHOME");
+        }
+        if (pythonHome != null) {
+            Pattern re = Pattern.compile("^python\\d\\d+\\.dll$");
+            for (File file : new File(pythonHome).listFiles()) {
+                if (!file.isFile()) {
+                    continue;
+                }
+                if (re.matcher(file.getName()).matches() && file.exists()) {
+                    System.load(file.getAbsolutePath());
                     return true;
                 }
             }


bsteffensmeier:
> Open questions:
> 
>     * Jep works if `python38.dll` is loaded in my case. There is also `python3.dll` there which does not work. Therefore proposed solution is looking for `pythonXX.dll` where XX is two digits minimum. What exactly corresponds to `libpython` on windows? Because `python38.dll` was found just by trial and error.

If windows is like linux then python38.dll and python3.dll are just links to the same file but the jep library will depend on one or the other. This would be determined at build time. Without digging into the build tools I think it is reasonable to assume the python build system would choose the more specific one so I think your assumption is fair.

>     * It was not tested for python2 and who knows how it will look in python4. So it might make sense to search for python3X.dll instead of pythonXX.dll since IIRC jep is going to drop support for python2 in future releases.

I think what you have is fine, we are specifically only doing it if there are problems so I don't see how we can make matters worse, if there are ways to do it better in python4 then hopefully someone can update this.

>     * There is an option to add base environment to `Path` during Anaconda installation. So it is possible that the user will set PYTHONHOME to the desired environment, but jep will load "wrong" python dll from the base environment. Maybe these attempts to load libpython from within PYTHONHOME should be performed by default before loading jep? And not only when jep fails with UnsatisfiedLinkError?

I am very worried about inadvertently breaking someone by trying to get involved earlier. People try to use jep in a variety of odd environments and I don't want to try to anticipate every possible case. My top priority is to allow things to work for people who configure their environment to be able to load the libraries correctly and jumping in earlier risks breaking those people.

pacher:
> I am very worried about inadvertently breaking someone by trying to get involved earlier. People try to use jep in a variety of odd environments and I don't want to try to anticipate every possible case. My top priority is to allow things to work for people who configure their environment to be able to load the libraries correctly and jumping in earlier risks breaking those people.

How about an option in `PyConfig`? Something like `setUsePythonFromPythonhome(true)` ?

bsteffensmeier:
> How about an option in `PyConfig`? Something like `setUsePythonFromPythonhome(true)` ?

Developers can load libpython from wherever they want before using jep with System.load. In my opinion that flexibility really makes it unnecessary for jep to add code to support everywhere libpython could be.

If you were to add such an option you need to ensure it works cross platform, it looks like sometimes libpython is in the root of PYTHONHOME and other times it is in lib, lib4 or Lib. You also need to handle the different library naming schemes, not only the various different extensions(especially macOS) but also the name variants described in PEP 3149. I also think it is common to install multiple python versions to the same location, for example on my ubuntu system libpython2.7.so is in the same directory as libpython3.6m.so so we would need some mechanism to handle that situation. All of these things are really platform dependent which is why I prefer to let the platform supported dynamic linker handle as much library loading as possible.

pacher:
Ok, last try to explain myself, I shall not bother you again with this


> My top priority is to allow things to work for people who configure their environment to be able to load the libraries correctly

I am trying to be these people. I have my local and isolated, carefully configured and curated conda environments. But instead of my carefully configured python, by default jep will load just whatever it finds in the PATH of the big system.
There are ways to deal with it, but as a developer I prefer bullet-proof solutions in code instead of administrative exercises.

> Developers can load libpython from wherever they want before using jep

Nope, not that easy. Exactly because of what you said next

> sometimes libpython is in the root of PYTHONHOME and other times it is in lib, lib4 or Lib

etc. ...
But wait, I've seen the code to deal this mess, it is in jep's `LibraryLocator.findPythonLibrary`, I just need to invoke it before starting jep!
But no, I can't. Because it is a private method of the class with private constructor used in the protected method `initialize` of the MainInterpreter.
So my options are:
1) Convince you to make it all public. Bad idea.
2) Use reflection. Similarly bad.
3) Copy - paste and maintain relevant code. Duplicating functionality, manual keeping up with upstream etc. Also bad.
4) Fork jep and implement it there. Slightly better than 3) but still bad.
5) Contribute a little option to the config class with minimal non-invasive changes to the existing code. Good idea.

And here we are. If I contribute the option, I can use it but everybody else can safely ignore it and don't even know about it. Config object is exactly the place to be bloated with this kind of stuff by design, negligible impact on API surface in my opinion.

Actually, now I think it will be much better to add a static method in LibraryLocator.  Something like `public static findPythonLibrary` just next to existing `public static findJepLibrary`. IMO looks consistent and fits perfectly. Just little addition with no changes to the existing code.

But well, If you don't want it, you have your reasons and I respect that. We don't have to discuss it further.

bsteffensmeier:
I appreciate the discussion. Loading libjep and libpython is by far the top source of questions about jep so if we can reduce confusion that would be great but I am worried if we introduce an option that don't work well it will increase confusion.

LibraryLocator.findPythonLibrary currently needs to know the name of the python library it is looking for. With your proposal to make it public I don't understand how anyone would determine the name of the python library in a portable way. The current behaviour relies on parsing the library name from the exception. This isn't reliable to begin with but you are suggesting people could call findPythonLibrary before the exception in which case we don't even have that message to try to parse.

My concern with making LibraryLocator public is really that it isn't actually very good at locating libraries. Having a public method for loading the correct libpython based off PYTHONHOME creates an expectation that we can successfully load libpython and I do not have any confidence that the current code in LibraryLocator can consistently do that on all platforms.

In my opinion there is no way we can solve this type of problem from Java. We don't even know the name of the file we need to load and it is going to be impossible to consistently load library dependencies when we don't know what they are. The exact filename for libpython is stored in the object header in libjep but from Java we have no way to get that critical piece of information. In my opinion that small hurdle already rules out any possibility of trying to solve all of our library loading problems from java.

Every operating system already provides the ability to perform dynamic linking based off the dependencies declared in libjep. In most systems it is very configurable. In Windows you can configure exactly which libraries will be loaded with the Path. If you have libraries that you don't want loaded on the Path then i don't think that is a problem with jep, I think that is a problem with your configuration.

pacher:
Alright, it is all fair points and I totally understand the hesitation. Just for the sake of discussion I will try to argue a little bit more.


> LibraryLocator.findPythonLibrary currently needs to know the name of the python library it is looking for. With your proposal to make it public I don't understand how anyone would determine the name of the python library in a portable way. The current behaviour relies on parsing the library name from the exception. This isn't reliable to begin with but you are suggesting people could call findPythonLibrary before the exception in which case we don't even have that message to try to parse.

True, but for example on Windows we don't even get the library name from exception. That is why in this PR I had to loop through the root of PYTHONHOME trying to find `pythonXX.dll` based on best guess. I suggest to have a similar search for `libpython*.*` in a few directories like lib and lib64.

The main assumption here is that there will be only one libpython in user-provided PYTHONHOME as opposed to your system directory with multiple different pythons. Do you have PYTHONHOME pointing to that directory? I am pretty new to python and only use miniconda local environments (also on linux server, which was highly recommended). If there is a use-case or any other reason to suggest the the user will have multiple libpython exactly in PYTHONHOME, then of course it all makes no sense and forget I asked. Once again, I only suggest to do it for the explicitly provided PYTHONHOME from the user, assuming the user knows what he/she is doing. I would restrict it to the value provided with PyConfig or as parameter and not even look into environment variable. And of course try my best to explain with method name and documentation that it is just a try on a best effort basis and probably throw a checked exception if none or multiple libpython found, to make it totally clear :)

Like in my case I anyway have to programmatically set pythonhome in PyConfig based on CONDA_PREFIX environment variable because otherwise something wrong is happening to `sys.path` in jep. So I could pass it to some API to load libpython before.
But you convinced me to just implement it again on my side.

PR #305, opened 2021-01-16
Search site packages for Mac framework builds

GavinDmello:
Thanks for this package.

I noticed that Jep does not search user site-packages for python framework builds, which leads to an Unsatisfied Link error. 

Using this fix would solve the issue.

### DIFF
--- a/src/main/java/jep/LibraryLocator.java
+++ b/src/main/java/jep/LibraryLocator.java
@@ -193,6 +193,32 @@ final class LibraryLocator {
             }
 
         }
+        
+        // For Mac framework builds
+        if (userHome != null) {
+            File localDir = new File(userHome, "Library");
+            if (localDir.isDirectory()) {
+                File pythonMainDir = new File(localDir, "Python");
+                if (pythonMainDir.isDirectory()) {
+                    for (File versionDir : pythonMainDir.listFiles()) {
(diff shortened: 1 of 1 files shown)


ndjensen (approved):

bsteffensmeier (approved):

bsteffensmeier:
I have merged this into the dev_4.0 branch for inclusion in the next release. Thanks for contributing.

tburch:
@bsteffensmeier any idea on when 4.0 will be released? If not, could this be folded into a point release since it doesn't seem like a breaking change?

bsteffensmeier:
I've been trying to do about a release per year, so the current plan was to release 4.0 no sooner than October 2021, a year after 3.9.1.

If the community needs a release before that I can do a release sooner. I'd like input from @ndjensen on whether he prefers a 3.9.2 with just this or all of 4.0.

tburch:
Sounds good. I would appreciate a sooner release if possible as this was kind of annoying when getting started with jep. Additionally,  setting `DYLD_LIBRARY_PATH` in the build configuration of a shared project is kind of troublesome.

ndjensen:
I don't think we should release a 3.9.2 with just this.  Maybe if more stuff comes up we could have a 3.9.2.  I'd prefer 4.0 but released sooner than October.

PR #308, opened 2021-01-28
[EXTREMELY WIP] Attempt to use the Python Stable ABI

Techcable:
The benefit of this is that jep could compile once and support multiple python versions. This could potentially allow the distribution of compiled binaries/wheels ahead of time.

Mostly this just means moving from fast access macros
to the checked runtime calls. This comes at some very minor performance cost (that is only present when we're compiling against the limited ABI).
Most of the changes so far involved the Unicode APIs. We can't have *direct* access to the internal `PyUnicode_KIND` representation.

Right now I've just hardcoded it on with a `#define PY_LIMITED_API` at the top of the `jep_platform.h` file, but of course I intend to make it configurable later.

It looks like some of the functions required for embedding are missing.

Not only are flags like `Py_VerboseFlag` and `Py_OptimizeFlag`
unavailable, so are things like `PyRun_String`.

We might have to sacrifice the `Py_VerboseFlag` and other config options
entirely, I'm not seeing a replacement available.
However, the other problems I'm sure I can work around....


Techcable:
SHOOOT! There was a git mixup. I'll rebase to master soon :laughing:

bsteffensmeier:
I recommend you work off the dev_4.0 branch so you don't have to work with all the python 2 compatibility.

Techcable:
Okay, I merged this onto the development branching the GitHub Web UI. Technically, this can't break the build because it wasn't working in the first place :see_no_evil: 

This is definitely *still unstable*

I want to bring up a couple of changes that I have made in this, some of which impact normal (non-stable API) compilations as well.
## Major changes so far
1. Using heap types for all extension types
    - In the python extension world, declaring Heap types are definitely less common than declaring type structures directly
    - This is absolutely needed for any version stability since `PyTypeObject` is opaque on the stable ABI and changes throughout versions
    - We really can't work around this, probably the only mandatory change to get this PR through.
2. Switch a lot of fast access macros like `PyList_GET_ITEM` to their function equivalents (like `PyList_GetItem`)
    - This should very rarely have a significant performance benefit. All it adds is bounds checks + type sanity.
    - This actually comes with the upside of better compatibility with PyPy and other alternative implementations.
3. Using new abstractions to emulate missing APIs
    - Some of the methods that JEP use are missing from the stable API. If I can't (or don't want to) remove their usages, I will wrap them in our own abstraction.
    - Unfortunately, many of the APIS relating to interpreter initialization, thread state, and sub interpreters are not in the stable API
        - Honestly, this makes sense given the impending 'GIL per sub interpreter' changes 
    - [ ] TODO: Remove Jep Buffer API
        - The abstraction I hacked together seems particularly ugly and probably will cause bugs for normal code down the line.
        - As far as I can tell the conversion logic for python buffers also handles `bytes` and `bytearray` objects, so I didn't want to remove it outright
        - It is probably far simpler to just change the conversion logic to fallback to different behavior on the stable API, than trying to emulate the entire buffer interface just for the benefit of  `bytes` and `bytearray`


Have a good day :smile:

src/main/c/Include/jep_platform.h
```diff
@@ -31,6 +31,18 @@
...
+ * If enabled Compile against the limited API, for version 3.8
+ *
+ * NOTE: This is higher than the minimum python version we normally require.
+ * However, the Limited API has been slowly expanded over time and we need to
+ * make use of the latest functionality. Almost all the "new" features in the Stable API
+ * are still present on old versions, only unstable.
+ *
+ * TODO: Make this configuration optional
+ */
+#define Py_LIMITED_API 0x03080000
```
Techcable, line 44: This is a hack.

Arguably, we could make this some sort of build flag given in the makfile or `setup.py`.
However, it's too unstable and feature-incomplete to ever become the default, so why bother?

.gitignore
```diff
@@ -25,3 +25,9 @@ dist
 MANIFEST
 .DS_Store
 .idea
+
+# Compilation database
+compile_commands.json
+
+# Clangd
+.cache/clangd
```
Techcable, line 33: Honestly, these should just be here anyways :wink:

src/main/c/Include/jep_platform.h
```diff
@@ -55,6 +67,80 @@ static_assert(PY_MAJOR_VERSION >= 3,"There is no Python 2 support!");
...
+    #undef PySequence_Fast_GET_SIZE
+    #undef PySequence_Fast_GET_ITEM
+    static inline Py_ssize_t PySequence_Fast_GET_SIZE(PyObject *target) {
+        if (PyList_Check(target) || PyTuple_Check(target)) {
+            return Py_SIZE(target); // see code
+        } else {
+            abort(); // Not a fast sequence: Undefined behavior
+        }
+    }
+    static inline PyObject *PySequence_Fast_GET_ITEM(PyObject *target, Py_ssize_t index) {
```
Techcable, line 119: Does the performance matter here?
If YES, then why are we paying the performance price of the stable ABI in other places? 
If NO, then why are we reimplementing this whole method here?

src/main/c/Include/jep_util.h
```diff
@@ -169,4 +169,6 @@ extern jclass JDOUBLE_ARRAY_TYPE;
 #define DEFINE_CLASS_GLOBAL(var, name) extern jclass var;
 CLASS_TABLE(DEFINE_CLASS_GLOBAL)
 
+int jep_util_type_ready(PyTypeObject **target_type, PyType_Spec *spec, PyTypeObject *base);
```
Techcable, line 172: This is a nice utility. I feel like in general, the project should switch to using heap-allocated types.

That seems like it's becoming the new sort of standard for CAPI extensions.

src/main/c/Include/pyembed.h
```diff
@@ -42,6 +42,27 @@ struct __JepThread {
...
 typedef struct __JepThread JepThread;
 
+#if defined(Py_LIMITED_API) && Py_LIMITED_API+0 < 0x03090000
+    /*
+     * On versions before 3.8, there was no way to get the current `PyInterpreterState*`
+     * using the Stable ABI.
+     *
+     * Therefore, we simply assume there is one and only one active interpreter (no sub-interpreters)
+     */
+    #define JEP_ASSUME_SINGLE_INTERPRETER
```
Techcable, line 52: This is an atrocious hack. Subinterpreters are an essential concept in `jep`.......

src/main/c/Include/pyembed.h
```diff
@@ -42,6 +42,27 @@ struct __JepThread {
...
+#endif
+
+#ifndef JEP_ASSUME_SINGLE_INTERPRETER
+static inline PyInterpreterState *pythread_get_interpreter(PyThreadState *state) {
+    #if Py_VERSION_HEX >= 0x03090000 || (defined(Py_LIMITED_API) && Py_LIMITED_API+0 >= 0x03090000)
+        return PyThreadState_GetInterpreter(state);
+    #elif !defined(Py_LIMITED_API)
+        return state->interp;
+    #else
+        #error "Can't go from PyThreadState -> PyInterpreterState before Py 3.9"
```
Techcable, line 62: So I forget here. Does the "Stable ABI"  support this use-case before Python 3.9?

src/main/c/Include/pyembed.h
```diff
@@ -72,4 +93,35 @@ JepThread* pyembed_get_jepthread(void);
...
+ * but works on the stable API too.
+ */
+PyObject *pyembed_util_run(JepRunTarget target, char* file_name, int input_type, PyObject *locals, PyObject *globals);
+static inline PyObject *pyembed_util_run_string(char *input_str, int input_type, PyObject *locals, PyObject *globals) {
+    JepRunTarget target = {
+            .target_type = JEP_RUN_STRING_C,
+            .c_string = input_str,
+    };
+    return pyembed_util_run(target, "<string>", input_type, locals, globals);
+}
```
Techcable, line 125: Reimplementing CAPI functionality on the stable ABI.

Yet another hack.....

src/main/c/Jep/convert_p2j.c
```diff
@@ -55,18 +55,25 @@ static void raiseTypeError(JNIEnv *env, PyObject *pyobject, jclass expectedType)
...
     if (PyJClass_Check(pyobject)) {
         actTypeName = "java.lang.Class";
     } else if (PyJObject_Check(pyobject)) {
         actTypeName = PyUnicode_AsUTF8(((PyJObject*) pyobject)->javaClassName);
     } else {
+#ifdef Py_LIMITED_API
+        name_object = PyObject_Str((PyObject*) Py_TYPE(pyobject));
+        actTypeName = PyUnicode_AsUTF8(name_object);
+#else
         actTypeName = pyobject->ob_type->tp_name;
```
Techcable, line 69: So there's no way to get `PyTypeObject->tp_name` in the Stable ABI, and I just approximate it as a sequence of Python-level calls.....

src/main/c/Jep/convert_p2j.c
```diff
@@ -729,9 +768,9 @@ static int pybufferview_check_jbytearray(Py_buffer view)
 static jbyteArray pybuffer_as_jbytearray(JNIEnv *env, PyObject *pybuffer)
 {
     jbyteArray jarr = NULL;
-    Py_buffer view;
+    JepBuffer view;
 
-    if (PyObject_GetBuffer(pybuffer, &view, PyBUF_FULL_RO) < 0) {
+    if (JepBuffer_FromObject_ReadOnly(pybuffer, &view) < 0) {
```
Techcable, line 773: Here is another API we end up emulating: The buffer API.

Turns out the buffer API provides a nice interface to dealing with bytes-like objects in general. We lose that when we use the standard API.

src/main/c/Jep/convert_p2j.c
```diff
@@ -745,28 +784,16 @@ static jbyteArray pybuffer_as_jbytearray(JNIEnv *env, PyObject *pybuffer)
...
-    }
-    if (PyBuffer_IsContiguous(&view, 'A')) {
-        (*env)->SetShortArrayRegion(env, jarr, 0, (jsize) length, (jshort*) view.buf);
-    } else {
-        void* buf = (*env)->GetShortArrayElements(env, jarr, NULL);
-        PyBuffer_ToContiguous(buf, &view, view.len, 'C');
-        (*env)->ReleaseShortArrayElements(env, jarr, buf, 0);
-    }
-    return jarr;
+    pybufferview_as_array_IMPL(jshort, Short)
```
Techcable, line 793: Actually, these macros are good because they clean up the code duplication.......

src/main/c/Jep/convert_p2j.c
```diff
@@ -702,24 +731,34 @@ static jobject PyCallable_As_JPyCallable(JNIEnv *env, PyObject *pyobject)
...
-        void* buf = (*env)->GetByteArrayElements(env, jarr, NULL);
-        PyBuffer_ToContiguous(buf, &view, view.len, 'C');
-        (*env)->ReleaseByteArrayElements(env, jarr, buf, 0);
-    }
+#ifdef Py_LIMITED_API
+    #define pybufferview_as_array_IMPL_COPY(jtype, Type) \
+        /* NOTE: Our limited wrappers only supported C */ \
+        (*env)->Set##Type##ArrayRegion(env, jarr, 0, (jsize) length, (jtype*) view.buf);
+#else
+    #define pybufferview_as_array_IMPL_COPY(jtype, Type) \
```
Techcable, line 739: Although I'm admittedly biased, I feel like these macros are a good thing in general because it significantly reduces code-duplication.

src/main/c/Jep/convert_p2j.c
```diff
@@ -929,28 +920,16 @@ static jfloatArray pybuffer_as_jfloatarray(JNIEnv *env, PyObject *pybuffer)
...
-    }
-    if (PyBuffer_IsContiguous(&view, 'A')) {
-        (*env)->SetDoubleArrayRegion(env, jarr, 0, (jsize) length, (jdouble*) view.buf);
-    } else {
-        void* buf = (*env)->GetDoubleArrayElements(env, jarr, NULL);
-        PyBuffer_ToContiguous(buf, &view, view.len, 'C');
-        (*env)->ReleaseDoubleArrayElements(env, jarr, buf, 0);
-    }
-    return jarr;
+    pybufferview_as_array_IMPL(jdouble, Double)
```
Techcable, line 929: So much deduplicated code.....

src/main/c/Jep/pyembed.c
```diff
@@ -65,11 +65,51 @@
     #endif
 #endif
 
-static PyThreadState *mainThreadState = NULL;
+typedef struct main_thread_state {
+    PyThreadState *py_state;
+#ifndef JEP_ASSUME_SINGLE_INTERPRETER
+    PyInterpreterState *py_interpreter;
```
Techcable, line 71: This is kind of gross. Used to emulate the incompleteness of the Stable ABI....

src/main/c/Jep/pyembed.c
```diff
@@ -206,6 +247,8 @@ void pyembed_preinit(JNIEnv *env,
                      jint hashRandomizationFlag,
                      jstring pythonHome)
 {
+    // TODO: Most of these flags can't be implemented with the Stable ABI :(
+#ifndef Py_LIMITED_API
```
Techcable, line 251: It seems like this is more proof we need more than what the Stable API can provide.

Python's CAPI interface is definitely a leaky abstraction, but we need the whole thing, warts and all.
We want to be able to control execution just as much as the `cpython` binary can......

For example the `Py_DontWriteBytecodeFlag` could be pretty important in restricted environments.....

src/main/c/Jep/pyembed.c
```diff
@@ -256,57 +300,48 @@ void pyembed_preinit(JNIEnv *env,
  */
 static int pyjtypes_ready(void)
 {
-    if (PyType_Ready(&PyJObject_Type) < 0) {
+    // start at the top with object
+    if (jep_jobject_type_ready() < 0) {
```
Techcable, line 304: I actually think this heap-type cleanup is a good idea even if we don't take this PR.

src/main/c/Jep/pyembed.c
```diff
@@ -551,7 +591,7 @@ void pyembed_shared_import(JNIEnv *env, jstring module)
 {
     const char *moduleName = NULL;
     PyObject   *pymodule   =  NULL;
-    PyEval_AcquireThread(mainThreadState);
+    PyEval_AcquireThread(get_main_pystate());
```
Techcable, line 594: This is gross. We're emulating yet another API on top of the Stable API......

src/main/c/Jep/pyembed.c
```diff
@@ -598,7 +638,20 @@ intptr_t pyembed_thread_init(JNIEnv *env, jobject cl, jobject caller,
...
+        /*
+         * We don't have direct access to the `PyInterpreterState*` object,
+         * se we hack around it by using the `PyGILState_Ensure()` API.
+         *
+         * That implicitly calls `PyGILState_Release` with the correct
+         * interpreter state ;)
+         */
+        PyGILState_STATE state = PyGILState_Ensure();
+        jepThread->tstate = PyThreadState_Get();
+        PyGILState_Release(state);
```
Techcable, line 651: How is this hack supposed to be more 'stable'?

src/main/c/Jep/pyembed.c
```diff
@@ -762,7 +821,15 @@ static PyObject* pyembed_jproxy(PyObject *self, PyObject *args)
 
     env = jepThread->env;
 
+#ifdef Py_LIMITED_API
+    inum = (int) PyList_Size(interfaces);
+    if (inum < 0) {
+        assert(PyErr_Occurred());
```
Techcable, line 827: What the heck?

src/main/c/Jep/pyembed.c
```diff
@@ -1357,12 +1431,118 @@ static void pyembed_run_pyc(JepThread *jepThread,
...
+            if (PyByteArray_Resize(buffer, used) != 0) {
+                goto finished_manual_run;
+            }
+            assert(PyByteArray_Size(buffer) == used);
+            char *result_string = PyByteArray_AsString(buffer);
+            if (result_string == NULL) goto finished_manual_run;
+            // Change target to UTF8 cstring, then delegate
+            target.target_type = JEP_RUN_STRING_C;
+            target.c_string = result_string;
+            exec_result = pyembed_util_run(target, file_name, input_type, locals, globals);
```
Techcable, line 1533: Lets not reimplement functionality that's already in CPython!!!

src/main/c/Objects/pyjobject.c
```diff
@@ -516,49 +525,31 @@ static PyMethodDef pyjobject_methods[] = {
 
 static PyMemberDef pyjobject_members[] = {
     {"__dict__", T_OBJECT, offsetof(PyJObject, attr), READONLY},
+    {"__dictoffset__", T_PYSSIZET, offsetof(PyJObject, attr), READONLY}, // NOTE: Sets `PyTypeObject.tp_dictoffset` for heap types...
```
Techcable, line 528: This is an interesting replacement for `tp_dictoffset`......
Looks like this is actually used in the internals somewhere...

Oddly, the heap-type API is mostly undocumented....

Techcable (commented):
I don't think this is a very good idea, because of the limitation of the Python CAPI.
It's not designed with embedding in mind, and `jep` is such a niche use case it's unlikely to support everything we need even if it becomes more popular....

Techcable:
I'm closing this because I feel the Python "Stable" CAPI is too limited and is unlikely to provide any benefits for `jep`.

I think some of the refactorings that I've done here is still a good idea, even though I don't think this PR should be accepted.
The switch to using heap-allocated types is debatable, but it seems like a good idea with the [ongoing encapsulation of the CAPI](https://www.python.org/dev/peps/pep-0620/). The PEP specifically identifies `PyTypeObject` as a particularly problematic structure for its version and implementation dependence.......

Thank you so much for your time!
I appreciate your effort maintaining jep :)

Issue #330, opened 2021-06-07
Override __repr__ of PyJObject


jsnps:
Hey guys, 

is it possible to override the __repr__ function of PyJObject (python wrappers around java objects)? 

In Jython there was a default __repr__ implementation, but an overriden toString method of PyObject subclasses was also taken as __repr__ override. We have quite some tests relying on these strings (e.g. a list of objects prints the repr string rather than the str string) - we admit that these tests should be reworked and not rely on the string representation of the objects. Another place, where these repr strings are appearing "unfamiliar" is our interactive console - it was easier to use / understand with the toString of our API objects. 
I experimented a bit and couldn't find a way to achieve this without going into the C code. Maybe I missed something?

I tried:
a) define a __repr__ function on the java objects, with the hope that this one could be picked up.
b) monkeypatch the function on the python object. I can "get" the __repr__ function, but when trying to assign something (eg. __str__) to it, python complains with "object has no attribute '__repr__'."

Is there any simple way to achieve this without modifying jep?

One idea (requiring jep modifications) would be to simply implement tp_repr - similar like tp_str, just checking for the existence of a repr method on the java object and use this one, if present. 
[PyJObect](https://github.com/ninia/jep/blob/master/src/main/c/Objects/pyjobject.c#L540)

Another idea (not thought through), implement the required functions for allowing to customize/extend PyJObject and provide a way to have control over what wrapper class should be instantiated. Maybe this ties into the custom converters idea, which was mentioned in another thread.

Thanks!

bsteffensmeier:
I can't think of any way to do it without modifying jep. I tried defining a `__repr__` method on the java object but it isn't picked up, I think the java methods are currently instance attributes instead of class attributes which may be why it doesn't work.

I would be in favor of modifying jep so that there is either a special name or attribute on a java method that could be detected by a custom tp_repr on pyjpbject to allow overrides.

ndjensen:
@bsteffensmeier, can we close this as included in Jep 4.0?

PR #339, opened 2021-07-19
Add __repr__ support for pyjobject. If method defined on java class, …

jsnps:
…it will be picked up for tp_repr of the python wrapper.

Hey, this is some suggestion for supporting custom __repr__ implementations in jep according to the discussion in https://github.com/ninia/jep/issues/330
Please review carefully, especially with regards to memory management in C ;)

For the test, I found test_object.py to be a good place, as it is also testing hash and toString, but I'm not sure about the naming conventions here (might be only meant for the Object class). I needed two simple java test classes (implementing __repr__), but none of the existing java Test* classes sounded like the right place for them, so I added TestPyJObject. 

@bsteffensmeier 
I'm not sure about performance, do you see this method as performance critical? I saw in jep 4.0 you are creating dedicated py types for all java classes https://github.com/ninia/jep/commit/151c748f6c3674bc1ef292156fa82224a208fa3c
Would this be a better place?

### DIFF
--- a/src/main/c/Objects/pyjobject.c
+++ b/src/main/c/Objects/pyjobject.c
@@ -272,6 +272,45 @@ static PyObject* pyjobject_str(PyJObject *self)
     return pyres;
 }
 
+// call __repr__() on jobject. returns null on error.
+// expected to return new reference.
+static PyObject* pyjobject_repr(PyJObject *self)
+{
+    PyObject   *pyres   = NULL;
+    JNIEnv     *env;
+    jmethodID  __repr__ = 0;
+    jstring    jrepr    = NULL;
+
+    //check if __repr__ method exists on java object
+    if (self->object && self->clazz) {
+        env   = pyembed_get_env();
+
+        Py_BEGIN_ALLOW_THREADS
+        if (JNI_METHOD(__repr__, env, self->clazz, "__repr__", "()Ljava/lang/String;")) {
+            jrepr = (jstring) (*env)->CallObjectMethod(env, self->object, __repr__);
+        } else {
+            //method does not exist, clear exception
+            if ((*env)->ExceptionCheck(env)) {
+                (*env)->ExceptionClear(env);
+            }
+        }
+        Py_END_ALLOW_THREADS
+        if (process_java_exception(env)) {
+            return NULL;
+        }
+        if (jrepr != NULL) {
+            pyres = jstring_As_PyString(env, jrepr);
+        }
+        (*env)->DeleteLocalRef(env, jrepr);
+    }
+
+    if (pyres == NULL) {
+        //default (see: PyObject_Repr)
+        pyres =  PyUnicode_FromFormat("<%s object at %p>", Py_TYPE(self)->tp_name, self);
+    }
+    return pyres;
+}
+
 
 static PyObject* pyjobject_richcompare(PyJObject *self,
                                        PyObject *_other,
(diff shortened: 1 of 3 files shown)


bsteffensmeier:
> I'm not sure about performance, do you see this method as performance critical? 
I can't think of any reason someone would be calling repr in performance sensitive code. One of the downsides of being an intermediary library is we never know for sure what anyone is doing with it but I would assume we are pretty safe here. One thought I have that could improve the performance with a minor change would be if we check self->attrs to see if __repr__ is present before getting the JNI method. I have not measured it but I suspect a single dict lookup would be faster than the JNI method lookup, especially since JNI encourages you to cache methodIDs so it appears that lookup is not heavily optimized. We would still need to do the JNI method lookup for cases where that attribute exists but that would be a much less common case and we could optimize further if it becomes a problem.

> I saw in jep 4.0 you are creating dedicated py types for all java classes [151c748](https://github.com/ninia/jep/commit/151c748f6c3674bc1ef292156fa82224a208fa3c)
> Would this be a better place?

Before this PR can be merged it will have to target dev_4.0. We only merge new code into dev_X.Y branches and master is updated when we release. See the [contribution guidelines](https://github.com/ninia/jep/blob/master/.github/CONTRIBUTING.md) for a summary of our process.

I think this code will continue to work as is on dev_4.0. Theoretically it would be better if we could leave the default tp_repr implementation on PyObject and check if a class defines __repr__ when we make the type and only assign an alternative tp_repr for only the classes that define it. We may even be able to make a PyJMethod wrapping the __repr__ java method and assign it to __repr__ in the dict during type creation and then we wouldn't need any new code for tp_repr if python can just hook directly into the PyJMethod. If you don't want to figure that out then I would be fine with just defining a common tp_repr on pyjobject like you have and we can always change it in a later release if necessary.

jsnps:
Hey, I think your suggestions make totally sense and I will try to incorporate these. :) Thanks for the feedback!

bsteffensmeier:
@jsnps Are you good if I merge this into dev_4.0? I would like to do a release of jep 4.0 in the next couple weeks and was hoping we can get this in. I am fine merging it as is, although I might investigate if we can check attr for __repr__ since a dict lookup seems simpler than JNI method lookup. Since this works I would be willing to do that investigation after a merge.

jsnps:
@bsteffensmeier  sorry for the inactivity - almost forgot about it due to other tasks ;) 
I'm fine with getting this merged and picking this up later on again, but I also did some experiment according to your suggestion just now. 
Inside the pyjtype_get_new method, I'm checking for a __repr__ method, create a pyjmethod and add it to the type dict. I reverted the repr override in pyjobject. My test seems to pick up the added method on the type and invokes it BUT the pyargs are empty inside pyjmethod_call leading to:  
`Traceback (most recent call last):
  File "/localdev/janiks/jep_contributions/tmp/jep/src/test/python/test_object.py", line 18, in test_repr
    self.assertEquals(repr(TestPyJObject.ReprClass()), "ReprClass")
RuntimeError: Invalid number of arguments: 0, expected 1.
`

Python seems to handle type and instance methods differently (not passing self), the comment in pyjtype says: "In the
     * future consider placing pyjfields or pyjmethods in this dict
     * to simplify the structure of each object.
  "

Do you have some suggestion, how we could get the pyjmethod to work on the pyjtype for java instance methods (non static methods)? Or is this the wrong place and we should add this method to another dict during pyjobject_init? It feels right, to introspect this during the type creation, as this will not change between different instances.

bsteffensmeier:
So normal python methods are part of the type dict but when you get them through the internal get_attr you create a bound instance of the method that is tied to a particular instance of the type and has a reference to self. I have branches where I have tried to make it so jep can store all pyjmethods on the type object but it gets very complicated and it always ends up being too big of a change to merge into jep. I have an old branch where I had to make pyjmethod a descriptor so that when you access the object it can bind the method by creating a PyMethod that contains the pyjmethod. https://github.com/bsteffensmeier/jep/blob/pyjtype2/src/jep/pyjmethod.c#L773

So while it maybe possible to tie `__repr__` to the type I don't think it is going to make it into 4.0

jsnps:
ok thanks for the explanation. Then I'm fine to get it in like this for now and I will revisit this, once you added the descriptor support on pyjmethod.

PR #353, opened 2021-12-01
Allow empty kwargs in calls to PyObjects

jonathanindig:
Python libraries expect to be able to wrap arbitrary functions and call them later by passing in `*args` and `**kwargs`.

See the [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps) example code, or the rather popular [`tqdm` library](https://github.com/tqdm/tqdm/blob/fc69d5dcf578f7c7986fa76841a6b793f813df35/tqdm/utils.py#L132).

Unfortunately, attempting this with a PyObject leads to a "Keywords are not supported." error.

Here's a (very) silly example using the `DisableOnWriteError` wrapper from `tqdm`:

```
$ jep
>>> import tqdm
>>> from java.lang import System
>>> from tqdm.utils import DisableOnWriteError
>>> x = DisableOnWriteError(System.out, None)
>>> x.write(1)
Keywords are not supported.
```

This PR allows invocation with empty kwargs to pass without raising an error.

### DIFF
--- a/src/main/c/Objects/pyjconstructor.c
+++ b/src/main/c/Objects/pyjconstructor.c
@@ -123,7 +123,7 @@ static PyObject* pyjconstructor_call(PyJMethodObject *self, PyObject *args,
     jobject   obj  = NULL;
     PyObject *pobj = NULL;
 
-    if (keywords != NULL) {
+    if (keywords != NULL && PyDict_Size(keywords) > 0) {
         PyErr_Format(PyExc_TypeError, "Keywords are not supported.");
         return NULL;
     }
--- a/src/main/c/Objects/pyjmethod.c
+++ b/src/main/c/Objects/pyjmethod.c
@@ -221,7 +221,7 @@ static PyObject* pyjmethod_call(PyJMethodObject *self,
     /* if params includes pyjarray instance */
     int            foundArray       = 0;
 
-    if (keywords != NULL) {
+    if (keywords != NULL && PyDict_Size(keywords) > 0) {
         PyErr_Format(PyExc_RuntimeError, "Keywords are not supported.");
         return NULL;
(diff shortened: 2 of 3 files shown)


bsteffensmeier:
This looks good to me. At this time I wasn't planning on doing another release on 4.0.x so I will create a new branch for dev_4.1 and merge this in there.

ndjensen (approved):

ndjensen:
I agree, @bsteffensmeier, with targeting Jep 4.1 for release, but if we happen to have a 4.0.1 we should cherry pick this into that release since this is such low risk.

jonathanindig:
Thanks for the quick reviews @ndjensen and @bsteffensmeier ! 

What is your typical release cadence for minor fixes such as this one? No pressure, just wondering whether we'll need to temporarily build this on our own internally as I have some users waiting on this fix. Thanks again!

bsteffensmeier:
> What is your typical release cadence for minor fixes such as this one? No pressure, just wondering whether we'll need to temporarily build this on our own internally as I have some users waiting on this fix. Thanks again!

I am currently only committing to a yearly release so nothing is planned until late 2022. If there is a strong need or alot of new code in then I would consider something earlier.

bsteffensmeier (approved):

jeremyrsmith:
@bsteffensmeier is there anything we can do to help with the release process? @jonathanindig and I are pretty reliant on jep at this point, and I think I speak for both of us by saying that we're happy to take on logistics tasks if that would be helpful.

ndjensen:
This is in the recently released Jep 4.0.1.

jonathanindig:
Thank you very much @ndjensen, much appreciated!

Issue #356, opened 2021-12-01
Update Jep import hooks to latest import hook API


ndjensen:
Jep has two import hooks, the Java importer and the Shared Modules importer.  According to [this documentation](https://docs.python.org/3/reference/import.html#the-meta-path) the method `find_module()` on import hooks is deprecated and will raise a warning in Python 3.10.  Therefore I tried it on Python 3.10 with the jep command line interpreter:
```
>>> import warnings
>>> warnings.filterwarnings("always")
>>> from java.util import HashMap
<frozen importlib._bootstrap>:914: ImportWarning: JepJavaImporter.find_spec() not found; falling back to find_module()
<frozen importlib._bootstrap>:671: ImportWarning: JepJavaImporter.exec_module() not found; falling back to load_module()
<frozen importlib._bootstrap>:914: ImportWarning: JepJavaImporter.find_spec() not found; falling back to find_module()
<frozen importlib._bootstrap>:671: ImportWarning: JepJavaImporter.exec_module() not found; falling back to load_module()
```
The warning did not appear until I turned warnings on.  Based on that output and [more documentation](https://docs.python.org/3/reference/import.html#loaders), the method load_module() is also deprecated.

Update both Jep custom importers to implement the newer API and therefore remove the deprecation warnings.

ndjensen:
This is part of the 4.0.1 release.

Issue #355, opened 2021-12-01
Update pyembed.c code to not have deprecation warnings on PyEval_InitThreads


ndjensen:
When building Jep 4.0 with Python 3.10, the following warning is produced:
```
src/main/c/Jep/pyembed.c: In function ‘pyembed_startup’:
src/main/c/Jep/pyembed.c:398:5: warning: ‘PyEval_InitThreads’ is deprecated [-Wdeprecated-declarations]
     PyEval_InitThreads();
     ^~~~~~~~~~~~~~~~~~
In file included from /usr/local/include/python3.10/Python.h:144,
                 from src/main/c/Include/jep_platform.h:35,
                 from src/main/c/Include/Jep.h:35,
                 from src/main/c/Jep/pyembed.c:44:
/usr/local/include/python3.10/ceval.h:122:37: note: declared here
 Py_DEPRECATED(3.9) PyAPI_FUNC(void) PyEval_InitThreads(void);
                                     ^~~~~~~~~~~~~~~~~~
```
The documentation also points this out: https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads

Update the code to not have the deprecation warning with newer versions of Python.  This may require the use of macros like `#IF PY_MINOR_VERSION > 8` depending on how old of versions of Python we're supporting, I'm not sure.

The documentation says this method will be removed in Python 3.11 so this should be targeted for Jep 4.1.0 at the latest or 4.0.x if there is an earlier release.

ndjensen:
Fix is here: https://github.com/ninia/jep/commit/0866d0a85554a0b9af372c549db571a66f524082

ndjensen:
This is part of the 4.0.1 release.

Issue #354, opened 2021-12-01
Replace build's distutils usage with setuptools


ndjensen:
When building Jep 4.0 with Python 3.10, we get the following warning:
```
DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils.core import setup, Extension
```

Philosophically, I had avoided depending on setuptools for the build because we didn't need it and I didn't want users to have to download an extra package/dependency.  However, that philosophy is superseded by Python's philosophy/attitude toward distutils and setuptools.  A nice thing about setuptools' independence from Python means it's not tied to the Python release schedule, so fixes and improvements can get in faster.

Switch the Jep build to use setuptools instead of distutils.  Ideally this will be targeted for Jep 4.1 so we will have completed this ticket well before the release of Python 3.12.

ndjensen:
Looks like [setuptools includes all of distutils](https://github.com/pypa/setuptools/commit/bb9fb1fcfe37c1ef1e29e1e6d1fc4e483c743380).  So the quick fix may be to change the imports throughout the files in the jep/commands directory.  A longer term fix may be to adapt Jep to a newer Python build system like the setup.cfg files.

ndjensen:
That did not work.
```
$ python3 setup.py clean
numpy not found, running without numpy support
running clean
Traceback (most recent call last):
  File "setup.py", line 139, in <module>
    'test': test,
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/setuptools/_distutils/core.py", line 148, in setup
    return run_commands(dist)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/setuptools/_distutils/core.py", line 163, in run_commands
    dist.run_commands()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/setuptools/_distutils/dist.py", line 967, in run_commands
    self.run_command(cmd)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/setuptools/_distutils/dist.py", line 984, in run_command
    cmd_obj = self.get_command_obj(command)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/setuptools/_distutils/dist.py", line 859, in get_command_obj
    cmd_obj = self.command_obj[command] = klass(self)
[...]

PR #357, opened 2021-12-01
Implement `isatty` method in StreamRedirect

jonathanindig:
The [`isatty()`](https://docs.python.org/3/library/io.html#io.IOBase.isatty) function can be used to check whether a file object is interactive.

By default, `sys.stdout` and `sys.stderr` implement this method, so some Python libraries assume it will always be present and call `sys.stdout.isatty()` without checking to see whether it's implemented.

Technically, at least as far as I am able to tell, there is no strict definition of what a Python "file object" is, so callers should really be checking whether the out file in question has the `isatty` attribute. Unfortunately, they often don't.

We ran into this issue [in Polynote](https://github.com/polynote/polynote/issues/696) and implemented `isatty` in our own wrapper. I figured it makes sense to implement it in your wrapper as well.

### DIFF
--- a/src/main/python/jep/redirect_streams.py
+++ b/src/main/python/jep/redirect_streams.py
@@ -40,6 +40,9 @@ class StreamRedirect(object):
     def flush(self):
         self.flushmethod()
 
+    def isatty(self):
+        return False
+
 def redirectStdout(javaOutputStream):
     sys.stdout = StreamRedirect(javaOutputStream)
 


bsteffensmeier (approved):

ndjensen (approved):
Thanks for the contribution!

jonathanindig:
Thank you for the review! If it's not too much to ask, it would be great to get this and #353 in some sort of release (4.0.1?) given their low risk. Is there anything I could do to help with that? Of course totally your call, just trying to avoid having to fork Jep internally if at all possible.

bsteffensmeier:
> If it's not too much to ask, it would be great to get this and #353 in some sort of release (4.0.1?) given their low risk. Is there anything I could do to help with that? Of course totally your call, just trying to avoid having to fork Jep internally if at all possible.

@ndjensen just identified several things that could cause us problems in python 3.11(#354, #355, #356). If those changes end up small I would like to get them in 4.0.1 also.

I'd also like to test out Java 17 before a release since that is the new LTS and if there are any issues there it would need another release.

I'll have more time at the end of the month so I am thinking of releasing 4.0.1 around new years if that helps you.

Issue #359, opened 2021-12-04
Tests freeze in python 3.9.9


bsteffensmeier:
When I try to build jep against python 3.9.9 the test hang and never complete. I have traced it to calling [sys.exit(0) in a test](https://github.com/ninia/jep/blob/master/src/test/python/subprocess/iter_itr_crash.py#L12). We may be able to work around this by removing that call from the test but I have also noticed sys.exit() will freeze an interactive interpreter, which might be a problem.

We have [some code for handling the exit but it just calls PyErr_PrintEx() to handle the exit](https://github.com/ninia/jep/blob/master/src/main/c/Jep/jep_exceptions.c#L61). I traced through PyErr_PrintEx() and found that the freeze occurs in t[hreading._shutdown()](https://github.com/python/cpython/blob/v3.9.9/Lib/threading.py#L1470). That function was recently changed in a [python bug](https://bugs.python.org/issue1596321) and I believe that change is what is causing the problem.

The problem does not exist in earlier versions of python 3.9. It does not exist in python 3.10.0 but the change was merged into the 3.10 branch so it will likely affect 3.10 whenever the next python version is released.

bsteffensmeier:
In #361 I have changed the behavior so that we will call exit() ourselves in native code rather than relying on python to exit. 

In the future I think we need to stop handling SystemExit specially and allow it to be wrapped as a JepException and handled in java. For Run.java we would like to catch that exception and exit from java but that will require a new mechanism for getting the exit code from a SystemExit wrapped in a JepException so there is more work to do.

ndjensen:
This is fixed in Jep 4.0.1.

PR #361, opened 2021-12-10
Fix handling of SystemExit to resolve freeze after cpython issue 1596321.

bsteffensmeier:


src/main/c/Jep/jep_exceptions.c
```diff
@@ -82,6 +72,34 @@ int process_py_exception(JNIEnv *env)
     PyErr_Fetch(&ptype, &pvalue, &ptrace);
 
     if (ptype) {
+        // let's not turn this into a Java exception if the user exited
+        if (ptype == PyExc_SystemExit) {
+            // If the exception is not normalized the exit code is in pvalue.
+            PyObject *code = pvalue;
+	    if (PyErr_GivenExceptionMatches(pvalue, ptype)) {
```
ndjensen, line 79: This line looks like it should be indented more.

src/main/c/Jep/jep_exceptions.c
```diff
@@ -82,6 +72,34 @@ int process_py_exception(JNIEnv *env)
...
+                 */
+                PyErr_Clear();
+            } else if (code == Py_None) {
+                exit(0);
+            } else if (PyLong_Check(code)) {
+                exit(PyLong_AsLong(code));
+            } else {
+                /*
+                 * It should be impossible to get here but if something changes
+                 * it seems reasonable to jsut let the SystemExit become a JepException
```
ndjensen, line 97: Spelled "just" wrong.

src/main/c/Jep/jep_exceptions.c
```diff
@@ -82,6 +72,34 @@ int process_py_exception(JNIEnv *env)
...
+                /* 
+                 * It should be impossible to get here unless the code attribute
+                 * is missing. There is no reasonable way to handle that case so
+                 * clear the error and let the SystemExit become a JepException
+                 */
+                PyErr_Clear();
+            } else if (code == Py_None) {
+                exit(0);
+            } else if (PyLong_Check(code)) {
+                exit(PyLong_AsLong(code));
```
ndjensen, line 93: I don't understand why you are calling exit.  That's not what we did before.

bsteffensmeier: That seems like the best way to maintain the current behavior. Previously we relied on PyErr_PrintEx which would eventually exit for us, since that will no longer reliably exits(see #359) it seems simplest to just exit ourselves.

ndjensen (approved):

PR #375, opened 2022-02-09
Fixed library location for Python 3.10.

node3112:
When checking for the Python library path, the pattern would not match more than one number.
This means that it jep would not be able to find the shared library for Python 3.10 installations.


bsteffensmeier (approved):
Looks good, thank you.

I think this change is minor enough to go in a 4.0.2 release, @ndjensen do you agree?

ndjensen (approved):
Agreed that this would be good in 4.0.2.  Thanks for the fix!

PR #380, opened 2022-03-08
Created a routine to search for .class files in directories in the cl…

nbanks:
Hello,

Before this change, I was unable to import one of my java classes from python in eclipse.  Eclipse compiles every .java file into a.class file, however it does not combine them into a jar.

The class loader can find my class.  This works:
Thread.currentThread().getContextClassLoader().getResource("my_class_name.class")

However the ClassList.loadPackages routine could only find my_class_name.class if I had instantiated this class in Java before I created the JEP SharedInterpreter.  This is because Package.getPackages() doesn't include this class until after it's been used once.  With this change, JEP searches for all .class files in the classpath as well as .jar files so I can import the Java class in Python even if it had never been used in Java.

I verified that JEP is using the same class loader object before making the change.

Our project has multiple packages which depend on each other, however I'm not sure if you need mulitple packages to recreate the problem.  Let me know if there's a more elegant solution you'd like me to test.

I'm using:
Eclipse 2021-09 (eclipse itself runs on java-11, but it calls the java 8 virtual machine)
Maven 3.6.3
Debian 11.2
adoptopenjdk-8-hotspot-amd64
Python 3.9

Thanks!

### DIFF
--- a/src/main/java/jep/ClassList.java
+++ b/src/main/java/jep/ClassList.java
@@ -75,8 +75,7 @@ public class ClassList implements ClassEnquirer {
     }
 
     /**
-     * load jar files from class path
-     * 
+     * load .jar files and .class files from class path
      */
     private void loadClassPath() {
         StringTokenizer tok = new StringTokenizer(
@@ -95,14 +94,21 @@ public class ClassList implements ClassEnquirer {
         while (!queue.isEmpty()) {
             String el = queue.remove();
 
-            if (!el.toLowerCase().endsWith(".jar")) {
-                // ignore filesystem classpath
+            // make sure it exists
+            File file = new File(el);
+
+            // A directory is a special case.
+            if (file.isDirectory()) {
+                // search for all .class files recursively starting with no prefix.
+                addClassFilesInTree(file, "");
                 continue;
             }
 
-            // make sure it exists
-            File file = new File(el);
             if (!file.exists() || !file.canRead())
+                continue;
+
+            // The .jar file is the normal case.
+            if (!el.toLowerCase().endsWith(".jar"))
                 continue;
 
             try (JarFile jfile = new JarFile(el, false)) {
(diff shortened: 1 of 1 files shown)


src/main/java/jep/ClassList.java
```diff
@@ -95,16 +94,23 @@ private void loadClassPath() {
...
+            if (file.isDirectory()) {
+                // search for all .class files recursively starting with no prefix.
+                addClassFilesInTree(file, "");
                 continue;
             }
 
-            // make sure it exists
-            File file = new File(el);
             if (!file.exists() || !file.canRead())
                 continue;
```
ndjensen, line 108: I think this check for file existence and readability should continue to be right after the File object is newed, and therefore before the file.isDirectory() check.

src/main/java/jep/ClassList.java
```diff
@@ -159,6 +165,31 @@ private void loadClassPath() {
         }
     }
 
+   /**
+     * Recursively go through a folder and all subdirectories looking for .class
+     * files. Add them all.
+     */
+    private void addClassFilesInTree(File folder, String prefix) {
```
ndjensen, line 172: Can you add javadoc `@param` to the javadoc on this method?  I realize ClassList.java is not very consistent with javadoc, but I'd like to improve the readabiity.  It took me a bit to figure out that prefix is going to be building a string of the package name as you walk down the directory tree, so maybe add a bit of explanation on that too?

src/main/java/jep/ClassList.java
```diff
@@ -159,6 +165,31 @@ private void loadClassPath() {
...
+                if (prefix!=null && !prefix.isEmpty()) {
+                    // Include a . between directories only if there was a prefix.
+                    addClassFilesInTree(file, prefix+"."+entry);
+                } else {
+                    // Don't include a prefix--we only care about subdirectories.
+                    addClassFilesInTree(file, entry);
+                }
+            } else if (file.exists() && file.canRead() && entry.toLowerCase().endsWith(".class")) {
+                // We've found a .class file on the file system.  Add it.
+                addClass(prefix, entry.replaceAll(".class$", ""));
```
ndjensen, line 188: Do you need the $ here?  I don't understand why that's there.

ndjensen (changes requested):
I think this is a good change, I can see how it will be helpful.  Just a couple of comments.

bsteffensmeier (approved):
The changes look good, thank you. I don't currently have plans for a 4.0.3 release so this will likely be merged into a 4.1 branch that I will need to create.

src/main/java/jep/ClassList.java
bsteffensmeier: That is the regex end of line character. I don't think there is a legal java filename with .class in the middle but it is pretty standard for ensuring you only modify the extension.

ndjensen: Thanks.

ndjensen:
A regression was found in the Jep 4.0 series on macOS so there will probably be a Jep 4.0.3 released in the next week or two with those fixes.  This change is low risk enough and doesn't change API that it can also go in 4.0.3.

ndjensen:
I will clean this up.

PR #386, opened 2022-03-27
Change Java fields and methods to type attributes instead of instance attributes

bsteffensmeier:
The relationship between PyJObjects and their types now more closely resembles
the behavior of normal Python types, allowing special methods such as __add__
and __repr__ to be picked up by Python when they are defined in Java classes.

Since all attributes of PyJObject have moved to the type the attr dict is no
longer necessary. However, every instance of PyJClass still shares a type so it
is not possible to move attributes of PyJClass to the type. Much of the code
for dealing with instance attributes has moved from PyJObject to PyJClass.

### DIFF
--- a/src/main/c/Jep/pyembed.c
+++ b/src/main/c/Jep/pyembed.c
@@ -182,11 +182,6 @@ static int initjep(JNIEnv *env, jboolean hasSharedModules)
             Py_DECREF(modjep);
             return -1;
         }
-        if (PyModule_AddObject(modjep, "__javaAttributeCache__", javaAttrCache)) {
-            Py_DECREF(javaAttrCache);
-            Py_DECREF(modjep);
-            return -1;
-        }
         PyObject *javaTypeCache = PyDict_New();
         if (!javaTypeCache) {
             Py_DECREF(modjep);
--- a/src/main/c/Jep/pyembed.c
+++ b/src/main/c/Jep/pyembed.c
@@ -177,11 +177,6 @@ static int initjep(JNIEnv *env, jboolean hasSharedModules)
         PyModule_AddStringConstant(modjep, "JCHAR_ID", "c");
         PyModule_AddStringConstant(modjep, "JBYTE_ID", "b");
         PyModule_AddIntConstant(modjep, "JEP_NUMPY_ENABLED", JEP_NUMPY_ENABLED);
-        PyObject *javaAttrCache = PyDict_New();
-        if (!javaAttrCache) {
-            Py_DECREF(modjep);
-            return -1;
-        }
         PyObject *javaTypeCache = PyDict_New();
         if (!javaTypeCache) {
             Py_DECREF(modjep);
--- a/src/main/c/Objects/pyjclass.c
+++ b/src/main/c/Objects/pyjclass.c
@@ -27,6 +27,11 @@
 
 #include "Jep.h"
 
+/*
+ * https://bugs.python.org/issue2897
+ * structmember.h must be included to use PyMemberDef
+ */
+#include "structmember.h"
 
 /*
  * Adds a single inner class as attributes to the pyjclass. This will check if
@@ -71,10 +76,7 @@ static PyObject* pyjclass_add_inner_class(JNIEnv *env, PyJClassObject *topClz,
         charName = jstring2char(env, shortName);
 
         if (PyDict_SetItemString(topClz->attr, charName, attrClz) != 0) {
-            printf("Error adding inner class %s\n", charName);
-        } else {
-            PyObject *pyname = PyUnicode_FromString(charName);
-            Py_DECREF(pyname);
+            return NULL;
         }
         Py_DECREF(attrClz); // parent class will hold the reference
         release_utf_char(env, shortName, charName);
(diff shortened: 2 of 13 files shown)


src/main/c/Objects/pyjclass.c
```diff
@@ -134,17 +136,63 @@ static PyObject* pyjclass_add_inner_classes(JNIEnv *env,
...
+ * Populate an attrs dict with all the Java methods and fields for the given
+ * class. This is currently the only way static methods and fields are made
+ * available. There is no known use for non-static fields but for now they are
+ * left in for backwards compatiblity, consider removing them in the future.
+ */
+static PyObject* pyjclass_init_attr(JNIEnv *env, jclass clazz)
+{
+    PyObject *attr = PyDict_New();
+    if (!attr) {
+        return attr;
```
ndjensen, line 149: Is the error indicator set by PyDict_New() in this probably-will-never-happen case?

src/main/c/Objects/pyjclass.c
```diff
@@ -134,17 +136,63 @@ static PyObject* pyjclass_add_inner_classes(JNIEnv *env,
...
+    }
+    PyTypeObject* type = PyJType_Get(env, clazz);
+    if (!type) {
+        Py_DecRef(attr);
+        return NULL;
+    }
+    PyObject *key, *value;
+    Py_ssize_t pos = 0;
+    while (PyDict_Next(type->tp_dict, &pos, &key, &value)) {
```
ndjensen, line 166: I don't understand how this while loop works.  Are all the fields and methods in tp_dict by this point?

src/main/c/Objects/pyjclass.c
```diff
@@ -276,12 +325,80 @@ static PyObject* pyjclass_call(PyJClassObject *self,
...
+// returns new reference.
+static PyObject* pyjclass_getattro(PyObject *obj, PyObject *name)
+{
+    PyObject *ret = PyObject_GenericGetAttr(obj, name);
+    if (ret == NULL) {
+        return NULL;
+    } else if (PyJMethod_Check(ret) || PyJMultiMethod_Check(ret)) {
+        /*
+         * TODO Should not bind non-static methods to pyjclass objects, but not
+         * sure yet how to handle multimethods and static methods.
```
ndjensen, line 339: So what happens at runtime with this change when you try to invoke an instance method on a pyjclass?

src/main/c/Objects/pyjtype.c
```diff
@@ -152,13 +180,130 @@ static PyObject* getBaseTypes(JNIEnv *env, PyObject *fqnToPyType, jclass clazz)
     return bases;
 }
 
+
+
+/*
+ * Adding empty slots to the type dict to prevent creation of __dict__ for
+ * every instance and ensures all attributes go through pyjfield.
```
ndjensen, line 187: Methods don't go through pyjfield, right?  I don't think this comment is right when it says "all attributes go through pyjfield".

src/main/c/Jep/pyembed.c
```diff
@@ -182,11 +182,6 @@ static int initjep(JNIEnv *env, jboolean hasSharedModules)
             Py_DECREF(modjep);
             return -1;
         }
-        if (PyModule_AddObject(modjep, "__javaAttributeCache__", javaAttrCache)) {
-            Py_DECREF(javaAttrCache);
-            Py_DECREF(modjep);
-            return -1;
-        }
         PyObject *javaTypeCache = PyDict_New();
```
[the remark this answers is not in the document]
bsteffensmeier: Oops, fixed.

src/main/c/Objects/pyjclass.c
bsteffensmeier: The documentation isn't clear on this. I looked at other places this is called in jep code and in cpython code and I can't find anywhere else the error indicator is set manually after this fails so I think the assumption is always that it is setting the error message.

src/main/c/Objects/pyjclass.c
bsteffensmeier: Yes, all the reflection happens before the type is created so we can just copy from the type. I added a comment to the function to clarify what it does.

src/main/c/Objects/pyjclass.c
bsteffensmeier: Before this change:
```
>>> from java.util import ArrayList
>>> ArrayList.size
<bound method size of <jep.PyJClass object at 0x7f88ebed50b0>>
>>> ArrayList.size()
Instantiate this class before calling an object method.
>>> instance = ArrayList()
>>> instance.size
<bound method size of <ArrayList object at 0x7f88ebed83b0>>
>>> instance.size()
0
>>> pyjtype = type(instance)
>>> pyjtype
<class 'java.util.ArrayList'>
>>> pyjtype.size
type object 'ArrayList' has no attribute 'size'
>>> pyjtype.size()
type object 'ArrayList' has no attribute 'size'
```
After this change:
```
>>> from java.util import ArrayList
>>> ArrayList.size
<bound method size of <jep.PyJClass object at 0x7f6231b97730>>
>>> ArrayList.size()
Instantiate this class before calling an object method.
>>> instance = ArrayList()
>>> instance.size
<bound method size of <java.util.ArrayList object at 0x7f6231ba0f80>>
>>> instance.size()
0
>>> pyjtype = type(instance)
>>> pyjtype
<class 'java.util.ArrayList'>
>>> pyjtype.size
<jep.PyJMethod object at 0x7f6231b9f8b0>
>>> pyjtype.size()
Invalid number of arguments: 0, expected 1.
>>> pyjtype.size(instance)
0
```
This particular comment was copied along with code from pyjobject, I think it is indicating pyjclass should behave like pyjtype and return unbound methods so you could pass in an instance. The comment I added above `pyjclass_init_attr` is suggesting that instead we should just remove them entirely from the pyjclass. The two comments are contradictory but I don't think we have a definitive path forward so I don't want to get rid of the old comment.

My rational for suggesting removing things from pyjclass is that eventually I think it would be nice if pyjclass was actually exposing the reflective methods from java.lang.Class, so from Python you could do some Java style reflection with things like getDeclaredMethods() and getAnnotations(). I think it would be confusing and complicated to have all the current methods of pyjclass mixed in the same object with those.

src/main/c/Objects/pyjtype.c
bsteffensmeier: Yes, I have updated the comment. I was looking at it only from the perspective of assigning values to attributes since that was a problem I was seeing but it is more clear to mention all the pyjthings.

src/main/c/Objects/pyjclass.c
```diff
@@ -141,6 +141,10 @@ static PyObject* pyjclass_add_inner_classes(JNIEnv *env,
  * class. This is currently the only way static methods and fields are made
  * available. There is no known use for non-static fields but for now they are
  * left in for backwards compatiblity, consider removing them in the future.
+ *
+ * Since the method and fields for a class are added to the type and the type
+ * is cached, this can just copy the attributes from he type and avoid any
```
ndjensen, line 146: he -> the

ndjensen (approved):

src/main/c/Objects/pyjclass.c
ndjensen: I like the new behavior, it behaves more like Python and is therefore better.

PR #395, opened 2022-05-04
Make most pyjtypes into heap types.

bsteffensmeier:
This brings jep closer to the Python Stable ABI and ensures the
pyjtypes with c extensions behave the same as the dynamically
created types generated for other classes.

See: https://docs.python.org/3.10/c-api/typeobj.html#static-types

### DIFF
--- a/src/main/c/Include/pyjautocloseable.h
+++ b/src/main/c/Include/pyjautocloseable.h
@@ -37,9 +37,6 @@
 #ifndef _Included_pyjautocloseable
 #define _Included_pyjautocloseable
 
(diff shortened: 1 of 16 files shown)


bsteffensmeier:
Thanks to @Techcable who provided a starting point for this work. This is not as extensive as the work in #308 but it is a step in that direction.

ndjensen (approved):

Techcable:
Great work! This should also make it easier to support alternative implementations like PyPy and GraalPython. 

Also the stable ABI has been gradually improving over the last couple of versions. I am hopeful #308 would go a lot better in the future :)

PR #396, opened 2022-05-06
Switch build to setuptools.

bsteffensmeier:
See #354. This change eliminates the DeprecationWarning during the build.

This does not remove every reference to distutils but I believe that any remaining references to distutils will continue to work after distutils is removed in Python 3.12 because of the distutils_hack in setuptools.

Many of the distutils references are related to choosing the c compiler in windows, since I do not have a local windows machine for testing I did not put any effort into windows. I did verify this change builds on windows on AppVeyor. I also have not tested on mac.

A few of the distutils references have open issues or fixes in newer versions of setuptools. I added comments to the setuptools issues and left the distutils code in for now so we are compatible with older versions of setuptools.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,7 +39,7 @@ Some benefits of embedding CPython in a JVM:
 
 Installation
 ------------
-Simply run ``pip install jep`` or download the source and run ``python setup.py build install``.
+Simply run ``pip install jep`` or download the source and run ``pip install .``.
 Building and installing require the JDK, Python, and optionally numpy to be installed beforehand.
 
 Dependencies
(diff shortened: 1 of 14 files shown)


ndjensen (approved):

ndjensen:
We need to add a pyproject.toml file that specifies we depend on setuptools.  See [PEP-518](https://peps.python.org/pep-0518/) and other Python projects.

ndjensen:
I just reread it and I guess we don't need a pyproject.toml file if we don't want one.  "Because the use of setuptools and wheel are so expansive in the community at the moment, build tools are expected to use the example configuration file above as their default semantics when a pyproject.toml file is not present."  The example configuration file has setuptools and wheel.

PR #402, opened 2022-05-24
Fixes #401 - issue NullPointerException in ClassList.loadClassList

### LINKED ISSUE #401: NullPointerException in ClassList.loadClassList when context class loader is unavailable in current thread.
**Describe the bug**
If we try to use Jep from a thread that has context class loader null (e.g. thread created from the outside of Java runtime) it fails with java.lang.NullPointerException at line  282 in ClassList in method loadClassList.

**To Reproduce**
Not yet have code to reproduce. A possible way similar to what we try to do:
- install Ant Media Server community
- git clone sample plugins https://github.com/ant-media/Plugins/ 
- open SamplePlugin with Eclipse JEE with Maven
- add Jep as dependency and try to instantiate it from f.ex. packet listener https://github.com/ant-media/Plugins/blob/master/SamplePlugin/src/main/java/io/antmedia/app/SamplePacketListener.java in onAudioPacket in a standard default way

**Expected behavior**
Instance of the class should be created without exception.

**Environment (please complete the following information):**
- OS Ubuntu 18.04
- Python 3.8
- OpenJDK Runtime Environment (build 11.0.15+10-Ubuntu-0ubuntu0.18.04.1)
- Jep Version: 4.0.3,  3.8.2
- Python packages: numpy

**Additional context**
When Thread.currentThread().getContextClassLoader() returns null then that null is added to the list of class loaders and then code tries to call in = cl.getResourceAsStream(rsc); and fails. A null check is needed before adding a class loader to the list.

Htarlov:
Fix #401 - for cases where Jep is run in an environment where context ClassLoader is unavailable and there is a null value.

### DIFF
--- a/src/main/java/jep/ClassList.java
+++ b/src/main/java/jep/ClassList.java
@@ -256,10 +256,17 @@ public class ClassList implements ClassEnquirer {
          * different ClassLoader than classes (e.g. tomcat), while the Jep.class
          * ClassLoader is useful if running inside an OSGi container as a Bundle
          * (e.g. eclipse).
+         * If there is no context ClassLoader then system ClassLoader will be tried.
          */
+        boolean hasContextClassloader = Thread.currentThread() != null
+                && Thread.currentThread().getContextClassLoader() != null;
+        ClassLoader parentClassLoader = hasContextClassloader
+                ? Thread.currentThread().getContextClassLoader()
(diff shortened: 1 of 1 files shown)


bsteffensmeier (approved):
Looks good, thanks for the fix.

ndjensen (approved):

PR #407, opened 2022-06-01
Implement custom MRO for PyJType.

bsteffensmeier:
Some Java classes have a type hierarchy that cannot be used with the Python C3 Method Resolution Order(MRO). This implements a custom MRO for PyJType so that it is possible to mirror the Java Class hierarchy in Python.

### DIFF
--- a/src/test/java/jep/test/TestPyJType.java
+++ b/src/test/java/jep/test/TestPyJType.java
@@ -1,0 +1,27 @@
+package jep.test;
+
+public class TestPyJType {
+
+    public static interface Root {
+    }
+
+    public static interface Child extends Root{
+    }
+
+    /**
+     * The default Python mro cannot create a consistent method resolution
+     * order (MRO) for this interface.
+     */
+    public static interface ProblemInterface extends Root, Child {
+    }
+
+    /**
+     * The default Python mro cannot create a consistent method resolution
+     * order (MRO) for this class
+     */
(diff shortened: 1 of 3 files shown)


src/test/java/jep/test/TestPyJType.java
```diff
@@ -0,0 +1,27 @@
+package jep.test;
+
+public class TestPyJType {
+
+    public static interface Root {
```
jeremyrsmith, line 5: Could it be worth adding some `default` methods here, to make sure that doesn't break stuff (now or in the future)?

bsteffensmeier:
> @bsteffensmeier I'm +1 to merge & release this, FWIW (notwithstanding my comment). It's certainly an improvement (and it fixes [polynote/polynote#1249](https://github.com/polynote/polynote/issues/1249) which would be great!)

Thank you. I plan to make a few more changes before this is merged. 

1. Ensure errors are checked and handled in the native code
2. Add a test case with default methods

It may take me a few weeks to find time to get back to this. 

I've only recently learned the significance of MRO and I am worried this change could have unintended side affects so I am not comfortable putting this in a 4.0.4 release. My plan is to merge this into the dev_4.1 branch for inclusion in the Jep 4.1 release, which will come out after Python 3.11 is released in October.

ndjensen (approved):

PR #421, opened 2022-08-24
SharedInterpreters now share MemoryManager and java PyObjects.

bsteffensmeier:
This enables all SharedInterpreters to use the same MemoryManager so that a PyPointer from any SharedInterpreter can be used on threads from any other SharedInterpreters. This requires internal changes to MemoryManager and the way it interacts with PyPointer because there is no longer a one-to-one relationship between a tstate and a MemoryManager, instead the same MemoryManager is valid for multiple tstates.

This should introduce more flexibility for developers. For ThreadPools performing the same task in different interpreters being able to reuse PyObjects should make them a more appealing feature.

I am hoping to expand this feature in a future release so that the MainInterpreter can clean up references from the shared MemoryManager, which would remove the requirement to keep an interpreter open by allowing cleanup without another Interpreter.

### DIFF
--- a/src/main/java/jep/python/PyPointer.java
+++ b/src/main/java/jep/python/PyPointer.java
@@ -75,8 +75,8 @@ public class PyPointer extends WeakReference<PyObject> {
      */
     protected synchronized void dispose() throws JepException {
         if (!disposed) {
-            decref(memoryManager.getThreadState(), pyObject);
             disposed = true;
+            decref(memoryManager.getThreadState(), pyObject);
             memoryManager.removeReference(this);
         }
     }
--- a/src/main/java/jep/python/PyPointer.java
+++ b/src/main/java/jep/python/PyPointer.java
@@ -38,8 +38,6 @@ import jep.JepException;
  */
 public class PyPointer extends WeakReference<PyObject> {
 
-    protected final long tstate;
-
     protected final long pyObject;
 
     protected final MemoryManager memoryManager;
@@ -53,17 +51,14 @@ public class PyPointer extends WeakReference<PyObject> {
      *            the PyObject (or subclass) corresponding to this PyPointer
      * @param memoryManager
      *            the MemoryManager responsible for tracking this pointer
-     * @param tstate
-     *            the pointer to the JepThreadState
      * @param pyObject
      *            the pointer to the PyObject*
      * @throws JepException
      *             if an error occurs
      */
     protected PyPointer(PyObject referrent, MemoryManager memoryManager,
-            long tstate, long pyObject) throws JepException {
+            long pyObject) throws JepException {
         super(referrent, memoryManager.getReferenceQueue());
-        this.tstate = tstate;
         this.pyObject = pyObject;
         this.memoryManager = memoryManager;
         this.memoryManager.addReference(this);
@@ -80,9 +75,9 @@ public class PyPointer extends WeakReference<PyObject> {
      */
     protected synchronized void dispose() throws JepException {
         if (!disposed) {
+            decref(memoryManager.getThreadState(), pyObject);
             disposed = true;
             memoryManager.removeReference(this);
-            decref(tstate, pyObject);
         }
     }
 
--- a/src/test/java/jep/test/TestSharedInterpreter.java
+++ b/src/test/java/jep/test/TestSharedInterpreter.java
@@ -106,10 +106,10 @@ public class TestSharedInterpreter extends Thread {
             t.join();
             String test2 = list.getAttr("pop", PyCallable.class).callAs(String.class);
             String test = list.getAttr("pop", PyCallable.class).callAs(String.class);
-            if (!test.equals("test")) {
+            if (!"test".equals(test)) {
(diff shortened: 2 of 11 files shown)


src/test/java/jep/test/TestSharedInterpreter.java
```diff
@@ -86,6 +92,53 @@ public static void testSharedTypes() throws Throwable{
...
+                    list.getAttr("append", PyCallable.class).call("test");
+                    interp2.set("l", list);
+                    interp2.exec("l.append('test2')");
+                }
+            });
+            t.start();
+            t.join();
+            String test2 = list.getAttr("pop", PyCallable.class).callAs(String.class);
+            String test = list.getAttr("pop", PyCallable.class).callAs(String.class);
+            if (!test.equals("test")) {
```
ndjensen, line 109: Best practice is to put the constant string on the left side of the .equals() call.  `!"test".equals(test)`

src/test/java/jep/test/TestSharedInterpreter.java
```diff
@@ -86,6 +92,53 @@ public static void testSharedTypes() throws Throwable{
...
+                throw new IllegalStateException("Expecting 'test', not " + test); 
+            }
+            if (!test2.equals("test2")) {
```
ndjensen, line 112: Best practice is to put the constant string on the left side of the .equals() call.  `!"test2".equals(test2)`

src/main/java/jep/python/PyPointer.java
```diff
@@ -80,9 +75,9 @@ protected PyPointer(PyObject referrent, MemoryManager memoryManager,
      */
     protected synchronized void dispose() throws JepException {
         if (!disposed) {
+            decref(memoryManager.getThreadState(), pyObject);
             disposed = true;
```
ndjensen, line 79: Why did you put the `decref` ahead of `disposed = true`?  If decref throws an exception then you could keep attempting to dispose it over and over again, which seems kinda bad.

src/test/java/jep/test/TestSharedInterpreter.java
```diff
@@ -86,6 +92,53 @@ public static void testSharedTypes() throws Throwable{
...
+            t = new Thread(() -> {
+                try {
+                    list.getAttr("append", PyCallable.class).call("test");
+                    pass[0] = false;
+                } catch (JepException e) {
+                }
+                try (Interpreter interp2 = new SubInterpreter()) {
+                    list.getAttr("append", PyCallable.class).call("test");
+                    pass[0] = false;
+                } catch (JepException e) {
```
ndjensen, line 126: What is the message on the exception that is thrown?  It's not clear to me how you made it work for SharedInterpreters but not SubInterpreters.

src/test/java/jep/test/TestSharedInterpreter.java
```diff
@@ -86,6 +92,53 @@ public static void testSharedTypes() throws Throwable{
...
+                }
+                try (Interpreter interp2 = new SubInterpreter()) {
+                    interp2.set("l", list);
+                    pass[0] = false;
+                } catch (JepException e) {
```
ndjensen, line 131: What is the message on the exception that is thrown?  It's not clear to me how you made it work for SharedInterpreters but not SubInterpreters.

ndjensen (approved):

src/main/java/jep/python/PyPointer.java
bsteffensmeier: I didn't consider that, I changed it to set the `disosed` before decref.

src/test/java/jep/test/TestSharedInterpreter.java
bsteffensmeier: Fixed

src/test/java/jep/test/TestSharedInterpreter.java
bsteffensmeier: Fixed

src/test/java/jep/test/TestSharedInterpreter.java
bsteffensmeier: The message is "Invalid thread access" from line 122 of MemoryManager.

The MemoryManager is keeping track of which interpreters are sharing memory and uses a ThreadLocal to track which threads those interpreters are valid on. When PyPointer or PyObject gets the thread state for any native operation it is now going through the MemoryManager which will return the thread state if there is an interpreter for that MemoryManager on that thread. If there is no interpreter for a memory manager then the "Invalid thread access" is thrown.

The implementation of the MemoryManager and PyPointer do not specifically behave differently for SubInterpreter vs. SharedInterpreter, But during construction the SharedInterpreters wil use a static MemroyManager that is shared and SubInterpreters will construct a new MemoryManager and this difference in construction causes different behavior.

src/test/java/jep/test/TestSharedInterpreter.java
bsteffensmeier: The message is "Invalid thread access" from line 122 of MemoryManager.

ndjensen (approved):

PR #440, opened 2022-11-05
Enable isinstance to work with PyJClass.

### LINKED ISSUE #438: Python isinstance does not work with imported Java classes
e.g.

```python
>>> from ghidra.program.database.data import *
...
>>> type(dt)
<class 'ghidra.program.database.data.FunctionDefinitionDB'>
>>> print(FunctionDefinitionDB)
class ghidra.program.database.data.FunctionDefinitionDB
>>> type(FunctionDefinitionDB)
<class 'jep.PyJClass'>
>>> isinstance(dt, FunctionDefinitionDB)
Traceback (most recent call last):
  File "/Users/pieceofsummer/.ghidra/.ghidra_10.1.5_PUBLIC/Extensions/Ghidrathon-1.0.0/data/python/jepeval.py", line 66, in jepeval
    more_input_needed = _jepeval(line)
  File "/Users/pieceofsummer/.ghidra/.ghidra_10.1.5_PUBLIC/Extensions/Ghidrathon-1.0.0/data/python/jepeval.py", line 49, in _jepeval
    exec(compile(line, "<string>", "single"), globals(), globals())
  File "<string>", line 1, in <module>
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

I see a workaround based on jep code found [here](https://github.com/ninia/jep/blob/b926ca4af1df7a496ea0a8cb7d030bbe310de898/src/test/python/test_import.py#L58-L59) that requires users to pass `__pytype__` e.g. `FunctionDefinitionDB.__pytype__`. Is there a way to use `isinstance` without a workaround?

See https://github.com/mandiant/Ghidrathon/issues/13 for more details.

bsteffensmeier:
This fixes #438

I'm not sure if this will go in 4.1.1 or 4.2, since we haven't created a dev_4.2 branch yet I am currently targeting dev_4.1 but I may wait to merge this until we decide on a release plan.

### DIFF
--- a/src/main/c/Objects/pyjclass.c
+++ b/src/main/c/Objects/pyjclass.c
@@ -392,12 +392,57 @@ static int pyjclass_setattro(PyObject *obj, PyObject *name, PyObject *v)
     return pyjfield_set((PyJFieldObject *) cur, pyjobj, v);
 }
 
-
 static PyTypeObject* pyjclass_GetPyType(PyJClassObject* self)
 {
     JNIEnv* env = pyembed_get_env();
     return PyJType_Get(env, self->clazz);
 }
+
+static PyObject* pyjclass_IsInstance(PyJClassObject* self, PyObject* inst)
+{
+    PyObject* type = (PyObject*) pyjclass_GetPyType(self);
+    if (!type) {
+        return NULL;
+    }
+    int result = PyObject_IsInstance(inst, type);
+    Py_DECREF(type);
+    return PyBool_FromLong(result);
+}
+
+static PyObject* pyjclass_IsSubclass(PyJClassObject* self, PyObject* sub)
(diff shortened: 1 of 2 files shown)


ndjensen (approved):

ctrueden:
This is great. To work around this limitation with the currently released version of Jep, I was doing `isinstance(obj, jtype.__pytype__)`; it'll be great to be able to use `isinstance(obj, jtype)` directly instead.

bsteffensmeier:
This will be included in the 4.1.1 release which I am planning to put out around the end of December.

ndjensen:
I forgot to mention, I think you should add an entry in the 4.1 release notes that points out isinstance now works as one might expect.

Issue #451, opened 2023-01-12
varargs support in overloaded methods


andvazqu:
pyjmultimethod won't match the correct method when using overloaded methods with variable arguments.

**To reproduce**
Use a java class with an overloaded method and variable arguments, such as:
![java_class](https://user-images.githubusercontent.com/48433864/212087745-81bf3087-afa8-4a5c-9fbb-5301147a70d0.png)

Calls to these methods from the jep console return "expected x but received y" or "No such method" errors:
![jep_console](https://user-images.githubusercontent.com/48433864/212088098-e17f29e5-7eee-4cba-9abf-0fd10bebf036.png)

**Expected behavior**
![jep_console_fix](https://user-images.githubusercontent.com/48433864/212088518-5b8855f7-eec9-480d-b7f1-68162146210f.png)

**Comments**
_pyjmultimethod_call_ checks _parameterCount == argsSize_, so a call to _method(1)_ is matched to _method(String...args)_ instead of to _method(int arg, String...args)_ .  _method()_ finds no compatible match.

Made some quick changes to _pyjmultimethod_call_  :
![changes_to_pyjmultimethocall](https://user-images.githubusercontent.com/48433864/212092188-1b43fa39-f92d-4f2f-bd67-1e7d28387efe.png)

and _PyJMethod_CheckArguments_ : 
![changes_to_pyjmethodcheckarguments](https://user-images.githubusercontent.com/48433864/212092218-e7cee089-4390-45aa-9151-9e0139578f3a.png)

to take into consideration variable arguments. This worked for my use case (see expected behavior), but haven't tested it thoroughly. 

**Environment:**
- Windows 10
- Python 3.9
- OpenJDK 11.0.15
- Jep Version 3.9.1 (saw same behavior on 4.1.1)

andvazqu:
Thanks for the quick reply! I added a type check for the use case you mentioned of a method overloaded with varargs of different types and created a pull request.

bsteffensmeier:
This PR was merged into the [4.2 release](https://github.com/ninia/jep/blob/v4.2.0/release_notes/4.2-notes.rst#improved-support-for-varargs) which was officially released today so I am closing this issue.

PR #452, opened 2023-01-19
Overloaded varargs

andvazqu:
See https://github.com/ninia/jep/issues/451

### DIFF
--- a/src/main/c/Objects/pyjmethod.c
+++ b/src/main/c/Objects/pyjmethod.c
@@ -187,6 +187,11 @@ int PyJMethod_CheckArguments(PyJMethodObject* method, JNIEnv *env,
         paramTypeId = get_jtype(env, paramType);
 
         match = pyarg_matches_jtype(env, param, paramType, paramTypeId);
+        if (match == 0 && varargs && paramTypeId == JARRAY_ID) {
+            jclass arrayType = java_lang_Class_getComponentType(env, paramType);
+            int arrayTypeId = get_jtype(env, arrayType);
+            match += pyarg_matches_jtype(env, param, arrayType, arrayTypeId);
+        }
         (*env)->DeleteLocalRef(env, paramType);
         if (PyErr_Occurred()) {
             matchTotal = -1;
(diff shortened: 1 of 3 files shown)


ndjensen:
This looks good to me.  Is it possible you could add a test to the end of test_call.py where the test fails without your fix and then passes with your fix?

src/main/c/Objects/pyjmethod.c
```diff
@@ -186,6 +187,11 @@ int PyJMethod_CheckArguments(PyJMethodObject* method, JNIEnv *env,
         paramTypeId = get_jtype(env, paramType);
 
         match = pyarg_matches_jtype(env, param, paramType, paramTypeId);
+        if (match == 0 && varargs && paramTypeId == JARRAY_ID) {
```
bsteffensmeier, line 190: This if statement needs to be limited to just the varargs parameter. If there is a method that takes an array and varargs then this will match the component type to the array.

For example this java method
```java
public String varargs(String[] arr, Object... varargs);
```
Would match this python call.
```python
obj.varargs("String")
```
Even though the call will fail.

bsteffensmeier (changes requested):
Thanks for putting this together. I have identified one edge case in an inline comment that should be fixed before this can be merged.

I agree unit tests would be nice, I  recommend test_overload instead of test_call.

bsteffensmeier:
The actual change in the jep code looks good to me. One of the unit tests is failing in AppVeyor in some environments for your test:

> Traceback (most recent call last):
>   File "C:\projects\jep\src\test\python\test_overload.py", line 86, in test_varargs
>     self.assertEqual(TestOverload.varargs(0), 'int...args')
> AssertionError: 'int i, String...args' != 'int...args'

I think that call is ambiguous and don't neither method seems more correct to me so in my opinion the test should pass if either method is used.

bsteffensmeier (approved):

bsteffensmeier:
I'll be merging this in to the dev_4.2 branch, which I have not created yet so this is waiting on me now. Thank you very much for the contribution.

PR #463, opened 2023-04-06
include pom files in built jars

ndjensen:
This improves traceability and may help some tools that introspect jars.

### DIFF
--- a/commands/java.py
+++ b/commands/java.py
@@ -269,13 +269,28 @@ class build_jar(Command):
                 os.makedirs(dest_dir)
             shutil.copy(src, dest)
 
-        self.spawn([self.jar, '-cf', 'build/java/jep-{0}-sources.jar'.format(self.version), '-C', 'build/java/jep.src/main/java', 'jep'])
-        self.spawn([self.jar, '-cfe', 'build/java/jep-{0}.jar'.format(self.version), 'jep.Run', '-C', 'build/java', 'jep'])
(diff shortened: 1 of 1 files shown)


commands/java.py
```diff
@@ -269,13 +269,28 @@ def build(self):
...
+
+    def setup_meta_inf(self, dest_dir):
+        pom_dir = os.path.join(dest_dir, 'META-INF', 'maven', 'black.ninia', 'jep')
+        if not os.path.exists(pom_dir):
+            os.makedirs(pom_dir)
+        shutil.copy('pom.xml', pom_dir)
+        pom_props ="""artifactId=jep
+groupId=black.ninia
+version="""
+        pom_props += self.version
```
bsteffensmeier, line 294: Does it matter that during development the version in pom.xml is not going to match the version in pom.properties? Right now on the dev_4.2 branch the [pom.xml has `4.2.0-SNAPSHOT`](https://github.com/ninia/jep/blob/dev_4.2/pom.xml#L9) and [version.py has just `4.2.0`](https://github.com/ninia/jep/blob/dev_4.2/src/main/python/jep/version.py#L26). When I do a release I remove the `-SNAPSHOT` from pom.xml so they will be matching in all released artifacts so I am fine if we want to ignore the mismatch during development.

bsteffensmeier (approved):

commands/java.py
ndjensen: I think it's fine.  I don't really want to parse the version out of pom.xml.

PR #465, opened 2023-04-09
Allow custom converters for Java To Python conversions.

bsteffensmeier:


src/main/c/Jep/convert_j2p.c
```diff
@@ -89,6 +89,35 @@ static PyObject* Character_As_PyObject(JNIEnv *env, jobject jobj)
     return jchar_As_PyObject(c);
 }
 
+/*
+ * This function calls user configurable conversion functions to convert a Java
+ * Object into a Python equivelant. The argument must be a PyJObject and if
```
ndjensen, line 94: equivalent

src/main/c/Jep/pyembed.c
```diff
@@ -132,6 +133,28 @@ static struct PyMethodDef jep_methods[] = {
         "to implement, string names])"
     },
 
+    {
+        "setJavaToPythonConverter",
+        pyembed_set_j2p_converter,
+        METH_VARARGS,
+        "Register a function for converting a Java Object of a specific Class to a Python equivelant.\n"
```
ndjensen, line 140: equivalent

ndjensen (approved):
Philosophical question: Should there be API on a Jep instance to set the Java to Python converter?  You would still have to get the PyCallable somehow but without it there will be a lot of code that is `interpreter.exec("jep.setJavaToPythonConverter(type, method)");` An `interpreter.setJavaToPythonConverter(Class, pycallable);` feels like it would be cleaner and more useful to me, cause usually (in my opinion) you'd probably want to set these up immediately on your Interpreter instance after instantiation.

bsteffensmeier:
> Philosophical question: Should there be API on a Jep instance to set the Java to Python converter? You would still have to get the PyCallable somehow but without it there will be a lot of code that is `interpreter.exec("jep.setJavaToPythonConverter(type, method)");` An `interpreter.setJavaToPythonConverter(Class, pycallable);` feels like it would be cleaner and more useful to me, cause usually (in my opinion) you'd probably want to set these up immediately on your Interpreter instance after instantiation.

I am really not sure the best way to expose the conversion registration. I was thinking that since the conversion function will usually be written in python that it is easiest to leave the object in python and register it there. 

If I understand your comment correctly you are worried that the current PR would cause a bunch of redundant code in this pattern:

```java
interpreter.exec("from java.package import JavaClassName");
interpreter.exec("def conversionFunction(javaThing):\n" +
                 "    return pythonThing");
interpreter.exec("jep.setJavaToPythonConversion(JavaClassName, conversionFunction)");
```
My biggest concern with having setJavaToPythonConversion in Java is that I can't think of any case where someone would already have a PyCallable in Java. So to retrieve a PyCallable and register in Java the code above would be rewritten as:

```java
interpreter.exec("def conversionFunction(javaThing):\n" +
                 "    return pythonThing");
PyCallable conversionFunction = interpreter.getValue("conversionFunction", PyCallable.class);
interpreter.setJavaToPythonConversion(JavaClassName.class, conversionFunction);
```
I don't think that is a significant improvement and I am not a fan of pulling Python functions into Java as PyCallable if the only use is to pass them back into Jep. 

For most projects I am thinking that there would be enough complex conversions that it wouldn't be done in Java with exec statements. Hopefully developers can put them all in a Python file and then just use an import statement or runScript to register all the conversion functions. In this case having the registration in python definitely seems simpler. I'm hoping to encourage something like this:

```python
import jep
from java.pacakge import classA, classB, classC

def conversionA(javaThing):
    return pythonThing

def conversionB(javaThing):
    return pythonThing

def conversionC(javaThing):
    return pythonThing

jep.setJavaToPythonConversion(ClassA, conversionA)
jep.setJavaToPythonConversion(ClassB, conversionB)
jep.setJavaToPythonConversion(ClassC, conversionC)
```

src/main/c/Jep/convert_j2p.c
bsteffensmeier: Fixed

src/main/c/Jep/pyembed.c
bsteffensmeier: Fixed

PR #467, opened 2023-04-20
Initial Implementation of PyBuiltins

bsteffensmeier:
This provides an interface to expose the functionality of the Python builtin module to Java using the Jep proxy functionality.

We have supported the proxy functionality for a long time but the only way for developers using Jep to use it is to write their own interface to use with the proxy. With this PR I am proposing that Jep can provide and maintain some interfaces for some standard Python objects to work with the Jep proxy functionality.

This PR exposes the builtin module because in some examples and tests I have written I have wanted a convenient way to access a few of these in Java and for others I can envision use cases where it might be helpful.

1. I think things like `type()` and `isinstance()` will be very helpful for type checking a `PyObject` in java.
2. `compile()`, `eval()` and `exec()` would be useful if anyone wants to build advanced interpreters beyond what Jep provides.
3. I included `dict()` because it would be a convenient way to get a globals /locals dict for exec().
4. I included the other collection types in case someone had a similar use cases where they needed to construct a Python collection directly.

I did not include functions that I didn't think would be particularly useful from Java, For example Java has it's own math library for doing `abs` and `pow`.

One of the important things to look at in the PR is the class of all the arguments and return types because often the proxy functionality would work equally well if we just used Object or PyObject everywhere. In places where certain types are required by the python functions I tried to be more specific and I tried to choose PyObject vs. Object based off what I thought made the most sense. The reason this is important is because if we need to change a return type or argument type in a future version of Jep then it could break backwards compatibility. Maintaining backwards compatibility is one of the major downsides to offering this type of support within Jep as opposed to requiring developers to create the interfaces themselves or offering them in a separate library.

It's also worth mentioning that everything PyBuiltins does could easily be done with `Interpreter.invoke()` or `Interpreter.getValue()` so this adds no new functionality to Jep, however I feel PyBuiltins provides a more object oriented way of accessing the functionality that will be more convenient for some Java developers using Jep.

### DIFF
--- a/src/main/java/jep/python/PyBuiltins.java
+++ b/src/main/java/jep/python/PyBuiltins.java
@@ -24,6 +24,8 @@
  */
 package jep.python;
 
+import java.util.Map;
+
 import jep.Interpreter;
 
 /**
--- a/src/main/java/jep/python/PyBuiltins.java
+++ b/src/main/java/jep/python/PyBuiltins.java
@@ -1,0 +1,98 @@
+/**
+ * Copyright (c) 2023 JEP AUTHORS.
+ *
+ * This file is licensed under the the zlib/libpng License.
+ *
+ * This software is provided 'as-is', without any express or implied
+ * warranty. In no event will the authors be held liable for any
+ * damages arising from the use of this software.
+ * 
+ * Permission is granted to anyone to use this software for any
+ * purpose, including commercial applications, and to alter it and
+ * redistribute it freely, subject to the following restrictions:
+ * 
+ *     1. The origin of this software must not be misrepresented; you
+ *     must not claim that you wrote the original software. If you use
+ *     this software in a product, an acknowledgment in the product
+ *     documentation would be appreciated but is not required.
+ * 
+ *     2. Altered source versions must be plainly marked as such, and
+ *     must not be misrepresented as being the original software.
+ * 
+ *     3. This notice may not be removed or altered from any source
+ *     distribution.
+ */
+package jep.python;
+
+import jep.Interpreter;
+
+/*
+ * Interface for accessing some of the functions of the Python builtins module from Java.
+ * More information on specific functions can be found at https://docs.python.org/3.5/library/functions.html
+ *
+ * To use this interface create an {@link Interpreter} and call {@link PyBuiltins#get(Interpreter)}.
+ *
+ * @since 4.2
+ */
+public interface PyBuiltins {
+
+    public boolean callable(Object object);
+
+    public PyObject compile(Object source, String filename, String mode);
+
+    public PyObject compile(Object source, String filename, String mode, int flags);
+
+    public PyObject compile(Object source, String filename, String mode, int flags, boolean dont_inherit);
+
+    public PyObject compile(Object source, String filename, String mode, int flags, boolean dont_inherit, int optimize);
+
+    public PyObject dict();
+
+    public void delattr(Object object, String name);
+
+    public String[] dir(Object object);
+
+    public Object eval(Object object, PyObject globals);
+
+    public Object eval(Object object, PyObject globals, PyObject locals);
+
+    public void exec(Object object, PyObject globals);
+
+    public void exec(Object object, PyObject globals, PyObject locals);
+
+    public PyObject frozenset();
+
+    public PyObject frozenset(Object iterable);
+
+    public Object getattr(Object object, String name);
+
+    public boolean hasattr(Object object, String name);
+
+    public long id(PyObject object);
+
+    public boolean isinstance(Object object, PyObject classinfo);
+
+    public boolean issubclass(PyObject cls, PyObject classinfo);
+
+    public PyObject list();
+
+    public PyObject list(Object iterable);
+
+    public PyObject object();
+
+    public PyObject set();
+
+    public PyObject set(Object iterable);
+
+    public void setattr(Object object, String name, Object value);
+
+    public PyObject tuple();
+
+    public PyObject tuple(Object iterable);
+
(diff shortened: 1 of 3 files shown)


src/main/java/jep/python/PyBuiltins.java
```diff
@@ -0,0 +1,98 @@
...
+ *     3. This notice may not be removed or altered from any source
+ *     distribution.
+ */
+package jep.python;
+
+import jep.Interpreter;
+
+/*
+ * Interface for accessing some of the functions of the Python builtins module from Java.
+ * More information on specific functions can be found at https://docs.python.org/3.5/library/functions.html
```
ndjensen, line 31: Why not link to the current documentation https://docs.python.org/3/library/functions.html?  Also will this generate javadoc with an html link in it, or do you need an `<a href="http://...">` tag?  See MemoryManager's javadoc.

src/main/java/jep/python/PyBuiltins.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public PyObject compile(Object source, String filename, String mode);
+
+    public PyObject compile(Object source, String filename, String mode, int flags);
+
+    public PyObject compile(Object source, String filename, String mode, int flags, boolean dont_inherit);
+
+    public PyObject compile(Object source, String filename, String mode, int flags, boolean dont_inherit, int optimize);
+
+    public PyObject dict();
```
ndjensen, line 49: It looks like you were going for alphabetical, which is good, but then dict() should be under delattr().

src/test/java/jep/test/TestPyBuiltins.java
```diff
@@ -0,0 +1,336 @@
...
+        if (!setBuiltin.equals(setGetValue)){
+            failure = "frozenset builtin does not return a full set";
+            return false;
+        }
+        return true;
+    }
+
+    private boolean testGetAttr() {
+        if (builtins.getattr(object, "__str__") == null){
+            failure = "hasattr builtin does not find __str__";
```
ndjensen, line 146: getattr, not hasattr.

src/test/java/jep/test/TestPyBuiltins.java
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    private boolean testSet() {
+        PyObject setBuiltin = builtins.set();
+        PyObject setGetValue = interp.getValue("set()", PyObject.class);
+        if (!setBuiltin.equals(setGetValue)){
+            failure = "set builtin does not return an empty set";
+            return false;
+        }
+        setBuiltin = builtins.set(new String[] {"1", "2", "3"});
```
ndjensen, line 232: Not a big deal, but a bit surprising that you support list(arg) and set(arg) but not dict(args).

src/test/java/jep/test/TestPyBuiltins.java
```diff
@@ -0,0 +1,336 @@
...
+            failure = "compile builtin does not return 1 + 1 = 2";
+            return false;
+        }
+        return true;
+    }
+
+    private boolean testDelAttr() {
+        subobject.setAttr("name", "value");
+        builtins.delattr(subobject, "name");
+        Set<String> dir = Set.of(builtins.dir(subobject));
```
ndjensen, line 67: It appears that `of` was introduced after Java 8, hence failing Appveyor.  I would prefer if we still supported Java 8.

src/test/java/jep/test/TestPyBuiltins.java
```diff
@@ -0,0 +1,336 @@
...
+        return true;
+    }
+
+    private boolean testType() {
+        if (!objectType.equals(builtins.type(object))){
+            failure = "type builtin returns wrong type for an object";
+            return false;
+        }
+        return true;
+    }
```
ndjensen, line 272: Your tests are good but they mostly test for expected behavior.  Should some testing test unexpected behavior, like you pass the wrong type of argument in and get an exception?  I'm ok with it if you don't want to do that.

ndjensen (changes requested):
Looks good but AppVeyor is failing and I found a test with an incorrect error message.

src/main/java/jep/python/PyBuiltins.java
bsteffensmeier: I have been working off of the 3.5 documentation because we don't want to expose any features that are not available in older versions of python that we still support, however given the limited number of functions we support and how annoying it would be to update when we no longer support old versions I have changed it to 3 and made it a proper link.

src/test/java/jep/test/TestPyBuiltins.java
bsteffensmeier: I just think it would confuse java developers to try to figure out what types of things to pass as an arg to a dict. I think it would have top be an iterable containing iterables of length 2 so like a List<List<?>> or an Object[][]. It seems like anyone who needs that and understands what should go into it would not be afraid to implement their own more specific proxy.

bsteffensmeier:
In addition to addressing the comments from @ndjensen I have also switched some of the arg types to be more specific classes. I have received feedback that the use of Object everywhere makes it hard to understand what type of java objects are actually OK to pass into the methods. My original justification for using Object is that it makes it more powerful to allow any Object that the jep conversion could make the right type. After further discussion I think it is more important that this class be something Java developers can understand easily. Advanced users who want to use more exotic conversions can easily extend the interface and overload any method to except other types of Objects. It would also be relatively simple for jep to change to less specific arguments in the future since it would remain compatible but switching to more specific args later would break things so this seems like a more conservative approach for an initial implementation,

ndjensen (approved):
Looks good.  I think it's easier to understand now.

PR #469, opened 2023-04-23
PyJType for java.util.Map now extends collections.abc.Mapping

bsteffensmeier:
This allows Java maps to be used in python anywhere a python mapping is expected. Before this change java Maps are missing functionality such as item(), keys(), and values() which are now automatically added by extending collections.abc.Mapping.


ndjensen (approved):
Looks good.

PR #472, opened 2023-05-09
Fixes for calling Java varargs constructors from Python.

bsteffensmeier:
I have been seeing intermittent test failures in some of the tests for `NDArray` and `DirectNDArray`. I found this was caused because overloaded constructors are using `pyjmultimethod` to resolve and since varargs support was added to `pyjmultimethod` the varargs constructor was sometimes chosen instead of te non-varargs constructor. Unfortunately `pyjconstructor` does not support varargs so this call would fail. I am not entirely sure why the problem is intermittent it may have to do with the order that the java compiler puts the constructors in the class files. The primary purpose of this commit is to add varargs support to `pyjconstructor`

Many of the changes are because in Java 8 both `java.lang.reflect.Method` and `java.lang.reflect.Constructor` extend `java.lang.reflect.Executable` so the java_access for these classes was consolidated where possible so that `pyjmultimethod` can call the appropriate methods to cover both methods and constructors.

The new logic in pyjconstructor to support varargs was borrowed from the existing logic in pyjmethod. Ideally this logic would be shared but there are some slight differences and alot of local variables so I didn't see a simple way to extract the shared logic.


src/main/c/Include/jep_util.h
```diff
@@ -133,6 +133,7 @@ extern jclass JDOUBLE_ARRAY_TYPE;
     F(JCHAR_OBJ_TYPE, "java/lang/Character") \
     F(JCHARBUFFER_TYPE, "java/nio/CharBuffer") \
     F(JNUMBER_TYPE, "java/lang/Number") \
+    F(JEXECUTABLE_TYPE, "java/lang/reflect/Executable") \
     F(JMEMBER_TYPE, "java/lang/reflect/Member") \
     F(JMETHOD_TYPE, "java/lang/reflect/Method") \
     F(JFIELD_TYPE, "java/lang/reflect/Field") \
```
ndjensen, line 139: Do you need to remove the java/lang/reflect/Constructor that is below here a few lines since you renamed the constructor C code to Executable?

src/main/c/Objects/pyjconstructor.c
```diff
@@ -113,27 +113,51 @@ int PyJConstructor_Check(PyObject* object)
...
+    /* The number of normal arguments before any varargs */
+    int            lenJArgsNormal   = 0;
+    PyObject      *firstArg         = NULL;
+    PyJObject     *clazz            = NULL;
+    int            pos              = 0;
+    jvalue        *jargs            = NULL;
+    /* if params includes pyjarray instance */
+    int            foundArray       = 0;
+    jobject        obj              = NULL;
+    PyObject     *pobj              = NULL;
```
ndjensen, line 128: Totally nitpicky, you can ignore this comment if desired, this last one *pobj is not aligned with the others (needs one more space).

src/main/c/Include/jep_util.h
bsteffensmeier: Fixed

src/main/c/Objects/pyjconstructor.c
bsteffensmeier: Fixed

ndjensen (approved):

PR #475, opened 2023-05-18
Create conversion functions for Java to numpy conversions.

bsteffensmeier:
Long term I hope to move most or all of the numpy specific logic to jep_numpy so that it is easier to disable/enable at runtime rather than compile time. This is a small step in that direction by removing the numpy specific logic in convert_j2p.c.

### DIFF
--- a/src/main/c/Jep/jep_numpy.c
+++ b/src/main/c/Jep/jep_numpy.c
@@ -199,25 +199,6 @@ int npy_array_check(PyObject *obj)
 
 
 /*
- * Checks if a jobject is an instance of a jep.NDArray
- *
- * @param env   the JNI environment
- * @param obj   the jobject to check
- *
- * @return true if it is an NDArray and jep was compiled with numpy support,
- *          otherwise false
- */
-int jndarray_check(JNIEnv *env, jobject obj)
-{
-    int ret = (*env)->IsInstanceOf(env, obj, JEP_NDARRAY_TYPE);
-    if (process_java_exception(env)) {
-        return JNI_FALSE;
-    }
-
-    return ret;
-}
-
-/*
  * Checks if a jobject is an instance of a jep.DirectNDArray
  *
  * @param env   the JNI environment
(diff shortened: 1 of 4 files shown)


src/main/c/Jep/jep_numpy.c
```diff
@@ -520,7 +501,8 @@ PyObject* convert_jdirectbuffer_pyndarray(JNIEnv *env, jobject jo, int ndims,
     return pyob;
 }
 
-PyObject* convert_jdndarray_pyndarray(JNIEnv *env, PyObject* pyobj)
+
+static PyObject* convert_jdndarray_pyndarray(PyObject* pyobj, PyObject* Py_UNUSED(unused))
```
ndjensen, line 505: What is the unused for?

src/main/c/Jep/jep_numpy.c
```diff
@@ -668,12 +652,12 @@ PyObject* convert_jprimitivearray_pyndarray(JNIEnv *env,
 /*
  * Converts a jep.NDArray to a numpy ndarray.
  *
- * @param env    the JNI environment
- * @param obj    the jep.NDArray to convert
+ * @param self   PyJObject representation of th NDArray
+ * @param unused Required by METH_NOARGS
```
ndjensen, line 656: I don't understand the unused.

ndjensen (approved):

src/main/c/Jep/jep_numpy.c
bsteffensmeier: Because this is used in a [PyMethodDef](https://docs.python.org/3/c-api/structures.html#c.PyMethodDef) it has to be a [PyCFunction](https://docs.python.org/3/c-api/structures.html#c.PyCFunction) which takes 2 args but since the flags indicate it is a [METH_NOARGS](https://docs.python.org/3/c-api/structures.html#METH_NOARGS) the second arg is unused.

PR #499, opened 2023-10-03
Add optional to allow execution script in different thread

ddovnar:
Hi! I'm ran into a problem with situation, when I trying to call multiple short scripts in the created singleton `Interpreter`.
To resolve this issue, I added optional flag to disable checking of `currentThread` in the `isValidThread` method.

### DIFF
--- a/src/main/java/jep/Jep.java
+++ b/src/main/java/jep/Jep.java
@@ -79,6 +79,8 @@ public abstract class Jep implements Interpreter {
 
     // windows requires this as unix newline...
     private static final String LINE_SEP = "\n";
+    
+    private boolean useCurrentThread = false;
 
     /**
      * Tracks if this thread has been used for an interpreter before. Using
@@ -212,6 +214,10 @@ public abstract class Jep implements Interpreter {
 
     private native long init(ClassLoader classloader, boolean hasSharedModules,
             boolean useSubinterpreter) throws JepException;
+    
+    public void setUseCurrentThread(boolean flag) {
+	this.useCurrentThread = flag;
+    }
 
     /**
      * Checks if the current thread is valid for the method call. All calls must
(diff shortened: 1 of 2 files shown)


bsteffensmeier:
1. This change will cause the same PyThreadState to be active on multiple threads. The [cpython documentation](https://docs.python.org/3/c-api/init.html#non-python-created-threads) indicates that each thread should create it's own PyThreadState so I don't think this is a supported way of using the API. My biggest concern is that cpython maintains a pointer to the current thread state and this is managed with the same API as the GIL. I am not confident the GIL will work as intended when reusing PyThreadState like this.
1. Have you done testing on this change using third party python extensions that use the [PyGILState_* C-API](https://docs.python.org/3/c-api/init.html#c.PyGILState_Ensure)? The reason for most of the threading constraints related to jep are because we have found cases where cpython will deadlock when PyGIlState_Ensure is called if there are multiple interpreters. This change seems to bypass all the protections we put in place to prevent deadlock so I am pretty sure there will be ways to deadlock in Jep with these changes. We will not accept this change if it re-introduces the deadlock problems we previously fixed with the threading constraints.

ddovnar:
Thank's for answer. This feature will give me what I need. It works in the default state as before, as it is only an additional parameter and may be optional (or experimental). Not all developers need to use it, but the main reason why it's needed in my case is that: I want to integrate java with python directly in the spring-cloud application. There are many ways to do such an integration, but I need use python as a micro-service in the spring-cloud ecosystem, and this feature will allow me not to worry about  workarround in the load-balancing and use python facilities closer to java application. How I use this, for example:

- Creating Spring service
- When a service is created, it creates SharedInterpreter (jep) and stores this object in the class member
- The Interpreter (jep) executes script and creates python object
- After all, any java code can call python object methods that are available in memory

PR #507, opened 2023-11-05
Java FunctionalInterfaces are automatically callable in Python

bsteffensmeier:
In Jep 3.7 we made it so that Python Callables can be converted to FunctionalInterfaces when passed to Java so I think it is appropriate to do the inverse when passing a FunctionalInterface to Python.

### DIFF
--- a/release_notes/4.2-notes.rst
+++ b/release_notes/4.2-notes.rst
@@ -31,3 +31,18 @@ Additions of PyBuiltins
 The `PyBuiltins <http://ninia.github.io/jep/javadoc/4.2/jep/python/PyBuiltin.html>`_
 class provides makes it easier to call some of the builtin Python functions
 from Java.
+
+Java FunctionalInterfaces are automatically callable in Python 
+**************************************************************
+When a Java object implementing a
+`FunctionalInterface <https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/FunctionalInterface.html>`_
+is used in Python it is automatically 
+`callable <https://docs.python.org/3/glossary.html#term-callable>`_
+and the `__call__` method will be set to the abstract method of the interface.
+Similar to the Java compiler, the type is not required to have the
+FunctionalInterface annotation to be treated as a functional interface.
+For example a Kava Function can be called directly without using the accept method:
+::
+    from java.util.function import Function
+    identity = Function.identity()
+    result = identity("abc") # Same behavior as identity.accept("abc")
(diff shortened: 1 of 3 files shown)


release_notes/4.2-notes.rst
```diff
@@ -31,3 +31,18 @@ Additions of PyBuiltins
...
+Java FunctionalInterfaces are automatically callable in Python 
+**************************************************************
+When a Java object implementing a
+`FunctionalInterface <https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/FunctionalInterface.html>`_
+is used in Python it is automatically 
+`callable <https://docs.python.org/3/glossary.html#term-callable>`_
+and the `__call__` method will be set to the abstract method of the interface.
+Similar to the Java compiler, the type is not required to have the
+FunctionalInterface annotation to be treated as a functional interface.
+For example a Kava Function can be called directly without using the accept method:
```
ndjensen, line 44: Did you mean Java where you wrote Kava?

src/main/c/Objects/pyjtype.c
```diff
@@ -271,6 +282,28 @@ static int addMethods(JNIEnv* env, PyObject* dict, jclass clazz)
...
+                Py_DECREF(pymethod);
+                return -1;
+            }
+            isAbstract = java_lang_reflect_Modifier_isAbstract(env, modifiers);
+            if (process_java_exception(env)) {
+                Py_DECREF(pymethod);
+                return -1;
+            }
+            if (isAbstract) {
+                if (oneAbstractPyJMethod) {
```
ndjensen, line 298: This if statement could use a comment, that it is checking if we had already had a method, in which case there were two methods and therefore it is not a functional interface.  It took me a bit to figure out how this was working, hence why I'd like a comment.

release_notes/4.2-notes.rst
bsteffensmeier: Yes thanks, fixed nnow. I am pretty sure I was correcting capitalization and inadvertently changed java to Kava.

src/main/c/Objects/pyjtype.c
bsteffensmeier: I added a comment. I was a little disappointed to be adding the extra complexity into this function, it is already doing a few different things I would have preferred to keep the functional interface more separate but I couldn't figure out how to do that without duplicating a bunch of this code and making the problem much harder so I opted to leave it mixed in with the other method creation. Anyway, comments are hopefully going to keep the complexity manageable so I appreciate you taking a fresh look at where they are needed.

ndjensen (approved):

PR #510, opened 2023-11-20
Java methods can accept Python kwargs.

bsteffensmeier:
The `@PyMethod` annotation can be used on a constructor or method in a Java class to customize the behavior of Jep when the method is called from Python. The most notable customization is the ability to pass kwargs from Python to Java. Jep will
automatically convert the kwargs to a PyObject or Map. It is still possible for a Java method to accept varargs from Python using Java style varargs syntax but the `@PyMethod` provides an additional way to specify a Java method accepts varargs that enables mixing varargs and kwargs. 

For example the following Java method signature shows a method that can be
called from Python using both varargs and kwargs:
::
    @PyMethod(varargs=true, kwargs=true)
    public void exampleMethod(Object[] varargs, Map<String,Object> kwargs) {
        // do stuff
    }


src/main/c/Objects/pyjmultimethod.c
```diff
@@ -137,12 +132,16 @@ static PyObject* pyjmultimethod_call(PyObject *multimethod,
         PyJMethodObject* method = (PyJMethodObject*) PyList_GetItem(mm->methodList,
                                   methodPosition);
         int parameterCount = PyJMethod_GetParameterCount(method, env);
-        jboolean varargs = java_lang_reflect_Executable_isVarArgs(env, method->rmethod);
-        if ((parameterCount == argsSize) || (varargs && argsSize >= parameterCount - 1)) {
+        if (keywords && PyDict_Size(keywords) > 0 && !method->isKwArgs) {
+            continue;
```
ndjensen, line 136: Why continue?  Doesn't this mean someone tried to pass kwargs to a method that does not support it?  Shouldn't we set the error indicator and abort?

src/main/java/jep/PyMethod.java
```diff
@@ -0,0 +1,60 @@
...
+     * varargs. The last argument(or second to last argument if kwargs is also
+     * true) must be a type that is compatible with a Python tuple, such as a
+     * List, array, or PyObject. If the argument is an array of a specific type
+     * then jep will attempt to convert all varargs to that type and throw an
+     * exception if conversion is not defined.
+     */
+    public boolean varargs() default false;
+
+    /**
+     * Set to true to indicate that this Java method should except keyword
```
ndjensen, line 55: except -> accept

src/main/c/Objects/pyjmethod.c
```diff
@@ -230,24 +268,19 @@ static PyObject* pyjmethod_call(PyJMethodObject *self,
...
-    }
     lenPyArgsGiven = PyTuple_Size(args);
 
     env = pyembed_get_env();
     lenJArgsExpected = PyJMethod_GetParameterCount(self, env);
     if (lenJArgsExpected == -1) {
         return NULL;
     }
+    if (self->isKwArgs) {
+        lenJArgsExpected -= 1;
```
ndjensen, line 279: I don't understand the +1 and -1 of len of args expected throughout this review.  Can you explain?

bsteffensmeier: lenJArgsExpected is how many regular python args need to be converted to java args. The easiest way to get the current code working with kwarrgs was to -1 here so it completely ignored the kwargs java expects and then add the kwargs at the end, That does make these variable names less accurate although I am having trouble coming up with a better name.

src/main/c/Objects/pyjmultimethod.c
bsteffensmeier: Someone tried to pass kwargs to a multi-method where one potential method in the multimethod doesn't support kwargs. Continue will check the next method in the multimethod, one of them might support kwargs.

src/main/java/jep/PyMethod.java
bsteffensmeier: fixed

src/main/c/Objects/pyjmethod.c
bsteffensmeier: In my latest change I have restructured the code so that when there are kwargs I decrement lenJArgsNormal instead of lenJArgsExpected, This really just means we +/- 1 in different places but overall I think it makes more sense because the kwargs are now included in expected but not included in normal. That makes sense to me because they are processed differently than other args so they are abnormal and expected. Let me know if you think the new way is worse, or if you have other ideas on how to make this cleaner.

src/main/c/Objects/pyjmethod.c
```diff
@@ -244,19 +244,41 @@ int PyJMethod_CheckArguments(PyJMethodObject* method, JNIEnv *env,
...
+ *
+ * This function primarily does two things.
+ *  1) Convert the python method arguments to java objects
+ *  2) Call a Java method
+ *
+ * Several different factors make the conversion of args complicated:
+ *  1) Python includes a reference to this/self as the first arg but Java needs
+ *     this/self separate from the other args.
+ *  2) Varargs are appended to the args by Python but Java varargs are a
+ *     seperate array.
```
ndjensen, line 261: seperate -> separate.  You spelled it right just above.

src/main/c/Objects/pyjmethod.c
```diff
@@ -244,19 +244,41 @@ int PyJMethod_CheckArguments(PyJMethodObject* method, JNIEnv *env,
...
+ *  3) Kwargs in Java are an ordinary arg but in Python they are passed
+ *     separatly.
```
ndjensen, line 263: separatly -> separately

ndjensen (approved):

src/main/c/Objects/pyjmethod.c
bsteffensmeier: Fixed

src/main/c/Objects/pyjmethod.c
bsteffensmeier: Fixed

PR #511, opened 2023-11-21
Adapt code to setuptools.dep_util deprecation.

qkaiser:
Installation of jep is broken with the recent deprecation of `setuptools.dep_util` in favor of `modified` (see https://github.com/pypa/setuptools/commit/0296279b68c7a29dbafd62f5c2d96220767bb4b6).

Tested with:
- [x] python 3.10
- [x] python 3.11
- [x] python 3.12

### DIFF
--- a/commands/scripts.py
+++ b/commands/scripts.py
@@ -11,7 +11,10 @@ import sysconfig
 
 from stat import ST_MODE
 from setuptools import Command
-from distutils.dep_util import newer
+try:
+    from setuptools.modified import newer
+except ImportError:
+    from distutils.dep_util import newer
 from commands.util import is_osx
 from commands.util import is_windows
 from commands.python import get_libpython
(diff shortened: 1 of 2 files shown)


qkaiser:
Ah ! Indeed it makes more sense to try importing the right way first before trying to import using the deprecated way :)

Adapted the code, let me know what you think.

commands/scripts.py
```diff
@@ -11,7 +11,10 @@
 
 from stat import ST_MODE
 from setuptools import Command
-from distutils.dep_util import newer
+try:
+    from distutils.modified import newer
```
bsteffensmeier, line 15: This is using distutils still, should this be setuptools also?

qkaiser: Done.

qkaiser: And yes, it should :)

bsteffensmeier (approved):

bsteffensmeier:
Thanks again, this will be included in the next release of jep which will come out before the end of the year.

PR #512, opened 2023-11-24
Support for isolated SubInterpreter

bsteffensmeier:
When using Python 3.12 or greater it is now possible to created isolated SubInterpreters which have separate GILs to improve concurrency. This feature can be enabled by setting the SubInterpreterOptions on JepConfig to SubInterpreterOptions.isolated(). For more details about how isolated sub-interpreters are different from ordinary interpreters refer to PEP-684.

### DIFF
--- a/src/main/java/jep/SubInterpreterOptions.java
+++ b/src/main/java/jep/SubInterpreterOptions.java
@@ -61,9 +61,9 @@ public class SubInterpreterOptions {
      * allocator state. Otherwise it will use (share) the main interpreter’s.
      * 
      * If this is false then check_multi_interp_extensions must be true. If this
-     * is true then gil must not be true.
+     * is true then own GIL must not be true.
      *
-     * Shared Modules cannot be used when this is set to true.
+     * Shared Modules cannot be used when this is set to false.
      *
      * @param useMainObmalloc
      *            whether the sub-interpreter will share the main interpreter's
@@ -77,8 +77,8 @@ public class SubInterpreterOptions {
     }
 
     /**
-     * If this is 0 then the runtime will not support forking the process in
-     * any thread where the sub-interpreter is currently active. Otherwise
+     * If this is false then the runtime will not support forking the process
+     * in any thread where the sub-interpreter is currently active. Otherwise
      * fork is unrestricted.
      *
      * Note that the subprocess module still works when fork is disallowed.
@@ -111,8 +111,8 @@ public class SubInterpreterOptions {
     }
 
     /**
-     * If this is false then the sub-interpreter’s threading module won’t
-     * create threads. Otherwise threads are allowed.
+     * If this is false then the sub-interpreter’s threading module won’t be
+     * able to create threads. Otherwise threads are allowed.
      *
      * @param allowThreads
      *            whether the  sub-interpreter will allow threads.
@@ -125,8 +125,8 @@ public class SubInterpreterOptions {
     }
 
     /**
-     * If this is false then the sub-interpreter’s threading module won’t
-     * create daemon threads. Otherwise daemon threads are allowed 
+     * If this is false then the sub-interpreter’s threading module won’t be
+     * able to create daemon threads. Otherwise daemon threads are allowed 
      * (as long as allowThreads is true).
      *
      * @param allowDaemonThreads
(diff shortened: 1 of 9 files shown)


src/main/java/jep/SubInterpreterOptions.java
```diff
@@ -0,0 +1,195 @@
...
+    protected SubInterpreterOptions(boolean isolated){
+        this.isolated = isolated;
+    }
+
+    /**
+     * If this is false then the sub-interpreter will use its own "object"
+     * allocator state. Otherwise it will use (share) the main interpreter’s.
+     * 
+     * If this is false then check_multi_interp_extensions must be true. If this
+     * is true then gil must not be true.
```
ndjensen, line 64: "then gil must not be true".  Do you mean "own gil"?  Also GIL should be capitalized.

src/main/java/jep/SubInterpreterOptions.java
```diff
@@ -0,0 +1,195 @@
...
+     * @return a reference to this SubInterpreterOptions
+     *
+     */
+    public SubInterpreterOptions setUseMainObmalloc(boolean useMainObmalloc) {
+        this.useMainObmalloc = useMainObmalloc ? 1 : 0;
+        return this;
+    }
+
+    /**
+     * If this is 0 then the runtime will not support forking the process in
```
ndjensen, line 80: I realize you probably pulled the wording straight from the Python docs, but I think since it's a boolean it should be "If this is true".

src/main/java/jep/SubInterpreterOptions.java
```diff
@@ -0,0 +1,195 @@
...
+     *
+     */
+    public SubInterpreterOptions setAllowExec(boolean allowExec) {
+        this.allowExec = allowExec ? 1 : 0;
+        return this;
+    }
+
+    /**
+     * If this is false then the sub-interpreter’s threading module won’t
+     * create threads. Otherwise threads are allowed.
```
ndjensen, line 115: "won't create threads" would sound a bit better as "won't be able to create threads".

src/main/java/jep/SubInterpreterOptions.java
```diff
@@ -0,0 +1,195 @@
...
+     *
+     */
+    public SubInterpreterOptions setAllowThreads(boolean allowThreads) {
+        this.allowThreads = allowThreads ? 1 : 0;
+        return this;
+    }
+
+    /**
+     * If this is false then the sub-interpreter’s threading module won’t
+     * create daemon threads. Otherwise daemon threads are allowed
```
ndjensen, line 129: "won't create dameon threads" would sound a bit better as "won't be able to create daemon threads".

src/main/java/jep/SubInterpreterOptions.java
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * This determines the operation of the GIL for the sub-interpreter. When
+     * this is false use (share) the main interpreter’s GIL. When this is true
+     * use the sub-interpreter’s own GIL.
+     *
+     * If this is true then useMainObmalloc must be false.
+     *
+     * @param ownGIL
+     *            whether the  sub-interpreter will use its own GIL.
```
ndjensen, line 168: double space in between "the  sub-interpreter"

ndjensen (approved):

src/main/java/jep/SubInterpreterOptions.java
bsteffensmeier: Fixed in 4f2db10

src/main/java/jep/SubInterpreterOptions.java
bsteffensmeier: Fixed in 4f2db10

src/main/java/jep/SubInterpreterOptions.java
bsteffensmeier: Fixed in 4f2db10

src/main/java/jep/SubInterpreterOptions.java
bsteffensmeier: Fixed in 4f2db10

ndjensen (approved):

Issue #538, opened 2024-06-17
`Py_DECREF` seems to be missing in `Java_jep_python_PyObject_setAttr`


CHanzyLazer:
**Describe the bug**
`Py_DECREF` seems to be missing in `Java_jep_python_PyObject_setAttr`, witch will cause memory leak when using `setAttr` of PyObject.

**To Reproduce**
```java
try (SharedInterpreter inter = new SharedInterpreter()) {
    inter.exec("class Simple:\n" +
               "    def __init__(self, id):\n" +
               "        self.member = None\n"+
               "        self.id = id\n" +
               "    def __del__(self):\n" +
               "        print('obj' + str(self.id) + ' del')"
    );
    PyCallable simple = inter.getValue("Simple", PyCallable.class);
    PyObject obj1 = simple.callAs(PyObject.class, 1);
    PyObject obj2 = simple.callAs(PyObject.class, 2);
    obj1.setAttr("member", obj2);
}
```

Output:

```
obj1 del
```

So only `obj1` was deleted, while `obj2` leaked.

**Expected behavior**

Output:

```
obj1 del
obj2 del
```

Both obj1 and obj2 should be deleted.

bsteffensmeier:
Thank you very much! I have created a PR with a fix, #539, please review it if you can. I plan to include the fix in the 4.2.1 release of Jep.

bsteffensmeier:
This has been included in the 4.2.1 release.

PR #539, opened 2024-06-17
Fix leaking attributes in java PyObject.

bsteffensmeier:
This is a fix for the problem identified in #538.

### DIFF
--- a/src/main/c/Jep/python/jep_object.c
+++ b/src/main/c/Jep/python/jep_object.c
@@ -109,12 +109,12 @@ JNIEXPORT void JNICALL Java_jep_python_PyObject_setAttr
         goto EXIT;
     }
     ret = PyObject_SetAttrString(pyObject, attrName, pyAttr);
-    Py_DECREF(pyAttr);
     if (ret == -1) {
         process_py_exception(env);
     }
 
 EXIT:
+    Py_XDECREF(pyAttr);
     PyEval_ReleaseThread(jepThread->tstate);
     release_utf_char(env, str, attrName);
 }
--- a/src/main/c/Jep/python/jep_object.c
+++ b/src/main/c/Jep/python/jep_object.c
@@ -109,6 +109,7 @@ JNIEXPORT void JNICALL Java_jep_python_PyObject_setAttr
         goto EXIT;
     }
     ret = PyObject_SetAttrString(pyObject, attrName, pyAttr);
+    Py_DECREF(pyAttr);
     if (ret == -1) {
         process_py_exception(env);
     }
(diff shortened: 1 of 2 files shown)


ndjensen (approved):

CHanzyLazer:
Wait, it seems necessary to check that `pyAttr` is not `NULL` before proceeding with `Py_DECREF(pyAttr);`, that is:
```C
    if (pyAttr) {
        Py_DECREF(pyAttr);
    }
```

CHanzyLazer:
> > Wait, it seems necessary to check that `pyAttr` is not `NULL` before proceeding with `Py_DECREF(pyAttr);`, that is:
> 
> jobject_as_PyObject() only returns null when the python exception is set so the check for process_py_exception() ensures that `pyAttr` is not null at that point.

Ok, so `Py_XDECREF(pyAttr);` should be under the `EXIT:` just like it is in `Java_jep_python_PyObject_getAttr`?

bsteffensmeier:
> > > Wait, it seems necessary to check that `pyAttr` is not `NULL` before proceeding with `Py_DECREF(pyAttr);`, that is:
> > 
> > 
> > jobject_as_PyObject() only returns null when the python exception is set so the check for process_py_exception() ensures that `pyAttr` is not null at that point.
> 
> Ok, so `Py_XDECREF(pyAttr);` should be under the `EXIT:` just like it is in `Java_jep_python_PyObject_getAttr`?

Anytime the python exception is set from jobject_as_PyObject() it will return null so in the case it does not hit the new DECREF it will be null and does not need a DECREF.

CHanzyLazer:
Okay, I'm not very familiar with Python's C interface. I was just considering whether we need to maintain consistency with the rest of the code, such as `Java_jep_python_PyObject_getAttr` and `Java_jep_python_PyObject_equals`.

bsteffensmeier:
> `Java_jep_python_PyObject_equals`



> Okay, I'm not very familiar with Python's C interface. I was just considering whether we need to maintain consistency with the rest of the code, such as `Java_jep_python_PyObject_getAttr` and `Java_jep_python_PyObject_equals`.

It looks to me like `Java_jep_python_PyObject_equals` is also leaking references. I will try to find some time to test and get a fix in as part of this patch, thanks for checking.

ndjensen:
I agree the `equals` method is also leaking, it's leaking `otherPyObject`.  Looks like I introduced both these leaks in Jep 3.8 when I reworked and improved the functionality of PyObject.  I am glad this was found and reported.  I checked the other methods in that file and they look ok with regards to memory.

bsteffensmeier:
I finally had some time to look at this closely. I added a test to see if equals is leaking and found it is not leaking, I just didn't notice the `Py_XDECREF` in the `EXIT` block, even though that is what @CHanzyLazer was trying to point out. I agree after reviewing the other functions in that file `Py_XDECREF` is the style there so I went ahead and switched the new DECREF over for consistency.

ndjensen:
Good catch.  I'm good with this going in.

PR #546, opened 2024-07-26
Auto detect JAVA_HOME on Linux

hahn80:
By tracking the link to javac we can find the actual path and set it to JAVA_HOME.


hahn80:
My black formatter seemed to make some unnecessary changes.

commands/java.py
```diff
@@ -48,6 +49,19 @@ def get_java_home():
         except CommandFailed:
             traceback.print_exc()
 
+    if is_linux():
+        try:
+            result = subprocess.run(
+                ["bash", "-c", "readlink -f $(which javac) | sed 's:/bin/javac::'"],
```
bsteffensmeier, line 55: This can be done in python without spawning a shell with `pathlib.Path(shutil.which("javac")).resolve().parents[1].as_posix()`. I would prefer we avoid spawning a shell unless we are doing something python can't easily do.

hahn80: That is a Pythonic way to do it. I will adjust the function based on your suggestion. Thanks for your review.

Makefile
```diff
@@ -0,0 +1,27 @@
...
+# Install target (optional, if you want to include it)
+install:
+	$(PYTHON) $(SETUP) install
+
+# Uninstall target (optional, if you want to include it)
+uninstall:
+	$(PYTHON) $(SETUP) uninstall
+
+.PHONY: all build clean install uninstall
+
```
ndjensen, line 27: I'd prefer if we didn't include the Makefile as I don't think @bsteffensmeier or myself will ever test or maintain it to make sure it still works.  Also I believe with newer versions of setuptools we're no longer supposed to call `python setup.py target` and this is one more thing we'd have to maintain as we shift away from that.

bsteffensmeier (approved):

ndjensen (changes requested):
Please remove the Makefile and .gitignore changes.

hahn80:
Thanks Nate.
I just removed the Makefile.
Hahn.

ndjensen (approved):

ndjensen:
Thanks for the contribution!

PR #551, opened 2024-08-26
Update threading checks to be compatible with Python 3.13

bsteffensmeier:
Python 3.13 is changing the way threading.active_count() works in sub-interpreters. In previous versions when no other threads in the interpreter are active then there is only 1 active thread. New in Python 3.13 the main thread if the main interpreter is included in the `active_count()` of sub-interpreters so there will always be 2 active threads. Since we check for active threads before ending the interpreter those checks are being updated for compatibility. The changes to Python that made this change necessary were made under python/cpython#114271. This problem was discovered in Python 3.13rc1 and the change has been verified in that version.

### DIFF
--- a/src/main/java/jep/Jep.java
+++ b/src/main/java/jep/Jep.java
@@ -438,17 +438,40 @@ public abstract class Jep implements Interpreter {
         }
 
         if (isSubInterpreter) {
-            exec("import sys");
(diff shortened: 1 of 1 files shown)


ndjensen (approved):
Looks good.

Issue #555, opened 2024-09-04
Memory Retention After Executing Python Function in SharedInterpreter


C3-BrijeshGujar:
**Describe the bug**
When using the Jep library, it appears that defining and executing a Python function within a SharedInterpreter instance causes memory to be retained, even after closing the interpreter. 

**To Reproduce**

1. Create a new SharedInterpreter instance and execute a Python function definition and invocation using the following code:
```java
jepInterpreter = new SharedInterpreter();
jepInterpreter.exec("def foo():\n\tpass\nfoo()");
```
2. Execute a subsequent memory-intensive operation:
```java
jepInterpreter.exec("x=[10000]*100000000");
```

Observe a memory spike in the JVM.

3. Close the interpreter:

```java
jepInterpreter.close();
```
Observe that the memory used by the interpreter is not released.

**Expected behavior**
The memory used by the interpreter should be released after the interpreter is closed.

**Environment :**
- OS Platform, Distribution, and Version: Fedora Linux 40 : Linux 6.8.11-300.fc40.x86_64
- Python Distribution and Version: 3.9
- Java Distribution and Version:  java-17-amazon-corretto
- Jep Version: 4.0.3

**Additional context**
Observed that this issue also happens on `jep==4.2.0`
If the line defining and invoking the function `jepInterpreter.exec("def foo():\n\tpass\nfoo()");` is removed, the memory is released as expected after closing the interpreter.

bsteffensmeier:
Thank you for providing clear instructions to easily recreate the problem.

The problem is that jep relies on Python reference counting to clear memory during `close()`. Python reference counting has a limitation that it cannot clear circular references. In this case the `foo()` function maintains a reference to the globals in `foo.__globals__` and the globals has a reference to foo which introduces a cycle that reference counting alone cannot free and since it involves globals everything in globals is kept in memory.

The good news is that the memory is not lost forever, if you have other SharedInterpreters running or if you open new SharedInterpreters then the Python garbage collection(gc) will eventually run and free the memory. I was able to free the memory by opening a new SharedInterpreter and running gc manually like this:

```java
try (SharedInterpreter interpreter = new SharedInterpreter()) {
    interpreter.exec("def foo():\n\tpass\nfoo()");
    interpreter.exec("x=[10000]*100000000");
}
Thread.sleep(30000); // observe the memory
try (SharedInterpreter interpreter = new SharedInterpreter()) {
    interpreter.exec("import gc");
    interpreter.exec("gc.collect()");
[...]

PR #556, opened 2024-09-04
Run Python garbage collector when an interpreter is closed.

bsteffensmeier:
This change will run Python Garbage Collection anytime an interpreter is closed.

#555 describes a simple use case where reference counting alone is unable to free memory in a jep interpreter on close and running the garbage collector cleans up the memory.

There is a potential downside that this will make close slower in some cases however it should be negligible in most cases and I believe the benefit of freeing memory promptly is worth the extra time in close().

### DIFF
--- a/src/main/c/Jep/pyembed.c
+++ b/src/main/c/Jep/pyembed.c
@@ -703,6 +703,7 @@ void pyembed_thread_close(JNIEnv *env, intptr_t _jepThread)
     Py_DECREF(key);
 
     Py_CLEAR(jepThread->globals);
+    PyGC_Collect();
 
     if (jepThread->classloader) {
         (*env)->DeleteGlobalRef(env, jepThread->classloader);


ndjensen (approved):
Looks good.

src/main/c/Jep/pyembed.c
```diff
@@ -703,6 +703,7 @@ void pyembed_thread_close(JNIEnv *env, intptr_t _jepThread)
     Py_DECREF(key);
 
     Py_CLEAR(jepThread->globals);
+    PyGC_Collect();
```
ndjensen, line 706: Just curious, why do you the gc collect call before the Py_EndInterpreter?

bsteffensmeier: It's not specifically documented for PyGC_Collect but my default assumption for any Python c-api call is that it requires the thread to have an interpreter or at least hold the GIL and after Py_EndInterpreter the thread has neither.

PR #562, opened 2024-09-16
Update include path for compatibility with numpy 2.

bsteffensmeier:
Numpy 2 is using a slightly different include path which is causing jep to build without numpy support even when numpy2 is installed. This change upgrades the build to find the new include path.


ndjensen (approved):
Looks good and what a quick turnaround!

PR #564, opened 2024-09-22
Ensure shared modules cannot be used with isolated interpreters.

bsteffensmeier:


src/main/java/jep/Jep.java
```diff
@@ -152,7 +152,7 @@ protected Jep(JepConfig config, boolean useSubInterpreter, MemoryManager memoryM
         boolean hasSharedModules = config.sharedModules != null
                 && !config.sharedModules.isEmpty();
 
-        if (hasSharedModules && config.subInterpOptions.useMainObmalloc == 0) {
+        if (hasSharedModules && (config.subInterpOptions.isolated || config.subInterpOptions.useMainObmalloc == 0)) {
             throw new JepException("Shared modules can only be used with a shared allocator.");
```
ndjensen, line 156: I kind of feel that the error message should be different if they used isolated, similar to the wording you have in the test.  But that could also lead into a mess of which error message to choose when they have both, or do you only give one error message at a time.  It's not a big deal so I will approve, it's more important that we have this safety check in there.

ndjensen (approved):

src/main/java/jep/Jep.java
bsteffensmeier: I have updated it to use two separate error messages. If they do both then the isolated interpreter error takes precedence.

Issue #323, opened 2021-04-23
Jars in manifest file not loaded and causes ModuleNotFoundError'>: No module named 'org.apache'


lfcnassif:
First, thank you all for this great project!
My app has the following structure:
./app.jar
./lib/lib1.jar
./lib/lib2.jar
...
My python script (called by java app) imports some classes from jars in lib folder. When executing `java -jar app.jar`, it throws jep.JepException: <class 'ModuleNotFoundError'>: No module named 'org.apache' for example. But all libraries are listed in app.jar manifest file and are used fine from app.jar embedded classes.

If I change current dir to lib folder and executes `java -jar ../app.jar` no error is thrown and the python script can access the java libraries. Default java classpath uses . directory.

Using jep-3.9.1 on Windows 10 with Oracle java 8 x64 and on Ubuntu 20.04 with Liberica OpenJDK 8 full.

lfcnassif:
Embedded manifest zipped below:

[MANIFEST.zip](https://github.com/ninia/jep/files/6368212/MANIFEST.zip)

bsteffensmeier:
Jep has the concept of a [ClassEnquirer](http://ninia.github.io/jep/javadoc/3.9/jep/ClassEnquirer.html) which tells the python envionment which classes are available in python. I suspect your problem lies with the default implementation. I am not very familiar with this code but it appears to be reading the manifest of every jar in the classpath and processing every library specified int he manifest so it seems like your case should work. I recommend debugging into the [ClassList](https://github.com/ninia/jep/blob/dev_3.9/src/main/java/jep/ClassList.java#L81) to try to determine if it is processing your manifest and finding the referenced libraries.

lfcnassif:
@bsteffensmeier thank you for the answer, will try to debug when some time is available and report back here.

reacan:
> If I change current dir to lib folder and executes `java -jar ../app.jar` no error is thrown and the python script can access the java libraries. Default java classpath uses . directory.

I had the exact same issue when your application with all the dependencies was set up in a Linux environment using a regular user account. When the application and all the dependencies were set up using the root account the application could be launched from any directory without an error. Hope this info will be of some help in case you haven't solved this issue yet.

lfcnassif:
@bsteffensmeier sorry for late reply, just had time to debug it. The issue is in this line https://github.com/ninia/jep/blob/dev_3.9/src/main/java/jep/ClassList.java#L120

[...]

PR #577, opened 2024-12-05
Fix ClassList to handle jar in working dir with classpath in manifest.

### LINKED ISSUE #576: Use jar runtime, ModuleNotFoundError, the reason is java.class.path to get the jar file getParent is null, resulting in the reference to the jar file path is not correct
Restoration results：
image

bsteffensmeier:
This fixes #576 which would come up if your class path includes a jar in the current working directory that has a class path in the manifest. The manifest processing assumes every class path entry has a parent dir(lib/foo.jar) but if the jar is in the current working directory then class path entries may not have a parent dir(foo.jar).

### DIFF
--- a/src/main/java/jep/ClassList.java
+++ b/src/main/java/jep/ClassList.java
@@ -127,8 +127,11 @@ public class ClassList implements ClassEnquirer {
                         String[] relativePaths = classpath.split(" ");
 
                         for (String relativePath : relativePaths) {
-                            String path = file.getParent() + File.separator
-                                    + relativePath;
+                            String path = relativePath;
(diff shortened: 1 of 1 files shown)


ndjensen (approved):

ndjensen:
This is the same area of code as mentioned in #323.  Based on the comment from June 1st, 2021, I think it also fixes that issue.  What do you think?

bsteffensmeier:
> This is the same area of code as mentioned in #323. Based on the comment from June 1st, 2021, I think it also fixes that issue. What do you think?

I agree this will also fix #323

PR #586, opened 2025-01-15
Update PyConfig to align with PEP 587.

bsteffensmeier:
The PyConfig Java class is currently used to set Python [Global Configuration Variables](https://docs.python.org/3/c-api/init.html#global-configuration-variables) but this part of the Python c-api has been deprecated since Python 3.12 and is scheduled for removal in 3.14. [PEP-587](https://peps.python.org/pep-0587/) describes a new mechanism for handling [Python Initialization Configuration](https://docs.python.org/3/c-api/init_config.html#python-initialization-configuration). This change removes the use of the deprecated variables and switches things over to the new way of doing things. This change should be fully backwards compatible for anyone using Jep, however many public methods in PyConfig are now deprecated.

Now that the Python c-api includes a PyConfig structure I changed our PyConfig class in Java to have fields and setters corresponding to the c-api structure. Making these match up keeps the native code simple and I think this will be easier to document and maintain moving forward but it makes this change confusing because [several of the fields were inverted](https://peps.python.org/pep-0587/#global-configuration-variables) in the transition. As an example there was previously a configuration variable called DontWriteBytecode and it has been replaced with a field called WriteBytecode so if previously DontWriteByteCode was true instead now WriteByteCode should be false. In the java PyConfig class this means the existing `setDontWriteBytecodeFlag()` method is deprecated and a new `setWriteBytecode()` has been added.

### DIFF
--- a/src/main/c/Include/pyembed.h
+++ b/src/main/c/Include/pyembed.h
@@ -43,7 +43,7 @@ struct __JepThread {
 typedef struct __JepThread JepThread;
 
 
-void pyembed_startup(JNIEnv*, jobjectArray, jint, jint,jstring, jint, jstring,
+void pyembed_startup(JNIEnv*, jobjectArray, jint, jint, jstring, jint, jstring,
                      jint, jint, jint, jint, jint);
 void pyembed_shutdown(JavaVM*);
 void pyembed_shared_import(JNIEnv*, jstring);
--- a/src/main/c/Include/pyembed.h
+++ b/src/main/c/Include/pyembed.h
@@ -43,9 +43,8 @@ struct __JepThread {
 typedef struct __JepThread JepThread;
 
 
-void pyembed_preinit(JNIEnv*, jint, jint, jint, jint, jint, jint, jint,
-                     jstring, jstring);
-void pyembed_startup(JNIEnv*, jobjectArray);
+void pyembed_startup(JNIEnv*, jobjectArray, jint, jint,jstring, jint, jstring,
+                     jint, jint, jint, jint, jint);
 void pyembed_shutdown(JavaVM*);
 void pyembed_shared_import(JNIEnv*, jstring);
 
--- a/src/main/java/jep/MainInterpreter.java
+++ b/src/main/java/jep/MainInterpreter.java
@@ -134,11 +134,11 @@ public final class MainInterpreter implements AutoCloseable {
             }
         }
 
-        if (pyConfig != null) {
-            setInitParams(pyConfig.noSiteFlag, pyConfig.noUserSiteDirectory,
-                    pyConfig.ignoreEnvironmentFlag, pyConfig.verboseFlag,
-                    pyConfig.optimizeFlag, pyConfig.dontWriteBytecodeFlag,
-                    pyConfig.hashRandomizationFlag, pyConfig.pythonHome, pyConfig.programName);
+        if (pyConfig == null) {
+            pyConfig = new PyConfig();
+        }
+        if (sharedModulesArgv != null) {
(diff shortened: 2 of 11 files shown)


src/main/c/Include/pyembed.h
```diff
@@ -43,9 +43,8 @@ struct __JepThread {
 typedef struct __JepThread JepThread;
 
 
-void pyembed_preinit(JNIEnv*, jint, jint, jint, jint, jint, jint, jint,
-                     jstring, jstring);
-void pyembed_startup(JNIEnv*, jobjectArray);
+void pyembed_startup(JNIEnv*, jobjectArray, jint, jint,jstring, jint, jstring,
```
ndjensen, line 46: Need to add a space between jint,jstring.

src/main/java/jep/MainInterpreter.java
```diff
@@ -258,7 +264,9 @@ public static void setInitParams(PyConfig config) throws IllegalStateException {
      *             if called after the Python interpreter is initialized
      * 
      * @since 3.7
+     * @deprecated Use {@link PyConfig#setArgv(String...)} instead.
```
ndjensen, line 267: How confident are you that this new alternative will have the same behavior?

src/main/java/jep/PyConfig.java
```diff
@@ -41,23 +54,195 @@ public class PyConfig {
...
+        this.useHashSeed = useHashSeed ? 1 : 0;
+        return this;
+    }
+
+    /**
+     * Set the default Python "home" directory, that is, the location of the
+     * standard Python libraries.
+     * 
+     * @param home
+     *            the directory tpuse for Python home.
```
ndjensen, line 129: Spelling "to use".

src/main/java/jep/PyConfig.java
```diff
@@ -67,9 +252,11 @@ public class PyConfig {
      * @param noSiteFlag
      *            value to pass to Python for Py_NoSiteFlag
      * @return a reference to this PyConfig
+     * @deprecated Use {@link #setSiteImport(int)} instead.
      */
+    @Deprecated
     public PyConfig setNoSiteFlag(int noSiteFlag) {
-        this.noSiteFlag = noSiteFlag;
+        this.siteImport = (noSiteFlag == 0) ? 1 : 0;
```
ndjensen, line 259: Mostly stylistic choice, but I think it might be slightly cleaner if you had these deprecated methods call the new non-deprecated methods to set the values.  Just a thought.

ndjensen (approved):

src/main/c/Include/pyembed.h
bsteffensmeier: Fixed

src/main/java/jep/PyConfig.java
bsteffensmeier: Fixed

src/main/java/jep/PyConfig.java
bsteffensmeier: I agree that is nicer, fixed.

src/main/java/jep/MainInterpreter.java
bsteffensmeier: For the primary goal of setting `sys.argv` for shared modules I am confident that the `TestSharedArgv` classes verifying the behavior is the same. I am not quite as confident that there are no other side effects. Both the old call to `PySys_SetArgvEx` and the new call to `PyConfig_SetBytesArgv` have documented optional side affects to set the path or parse the command line for options however in both cases we are setting it up to not do anything extra so as far as I can tell they are doing the same thing.

ndjensen (approved):
Looks good.

PR #592, opened 2025-01-23
Move build metadata from setup.py to pyproject.toml.

bsteffensmeier:
This moves the build metadata into pyproject.toml. As far as I can tell there is no immediate benefit to this change but it aligns with current best practices for building a python project and I suspect it is inevitable we will need to use pyproject.toml at some point in the future so I would like to have a starting version to work with in the future.

With this change I am still able to build using either `pip install --no-build-isolation .` or `setup.py install`. I still get a deprecation warning when running setup.py directly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta:__legacy__"
 name = "jep"
 description = "Jep embeds CPython in Java through JNI."
 dynamic = ["version"]
-license = { "text" = "zlib/libpng"}
+license = { "text" = "Zlib"}
 readme = "README.rst"
 requires-python = ">= 3.10"
 authors = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,6 @@ build-backend = "setuptools.build_meta:__legacy__"
 name = "jep"
 description = "Jep embeds CPython in Java through JNI."
 dynamic = ["version"]
-#version = "4.3.0"
 license = { "text" = "zlib/libpng"}
 readme = "README.rst"
 requires-python = ">= 3.10"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,45 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta:__legacy__"
+
+[project]
+name = "jep"
+description = "Jep embeds CPython in Java through JNI."
+dynamic = ["version"]
+#version = "4.3.0"
+license = { "text" = "zlib/libpng"}
+readme = "README.rst"
+requires-python = ">= 3.10"
+authors = [
+  {name = "Mike Johnson", email = "mike@mrj0.com"}
+]
+maintainers = [
+  {name = "Nate Jensen", email = "ndjensen@gmail.com"},
+  {name = "Ben Steffensmeier", email = "ben.steffensmeier@gmail.com"}
+]
+classifiers = [
+  "License :: OSI Approved :: zlib/libpng License",
+  "Development Status :: 5 - Production/Stable",
+  "Intended Audience :: Developers",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,45 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
+build-backend = "setuptools.build_meta:__legacy__"
+
+[project]
+name = "jep"
+description = "Jep embeds CPython in Java through JNI."
+dynamic = ["version"]
+#version = "4.3.0"
```
ndjensen, line 9: Why is version commented out?

pyproject.toml
```diff
@@ -0,0 +1,45 @@
+[build-system]
+requires = ["setuptools >= 61.0"]
```
ndjensen, line 2: Is this the version that first included distutils?  How did you pick the version?

ndjensen (commented):
How are you able to tell the install is using pyproject.toml instead of setup.py?  Or does it use both?

Do you know if there is still a way to run individual build commands like clean, java, build_ext, test?

bsteffensmeier:
> How are you able to tell the install is using pyproject.toml instead of setup.py? Or does it use both?

It is always using both.

When you run a `pip install` then pip will read pyproject.toml and the build-system section instructs it to use setuptools and setuptools has logic which calls setup.py. Then when `setuptools.setup()` runs it reads pyproject.toml and fills in all the metadata that was previously in python.

If you just call setup.py without pip then it still loads metadata from pyproject.toml.

> Do you know if there is still a way to run individual build commands like clean, java, build_ext, test?

I am able to use those by running setup.py directly. There is no change that I can see in how those work. setup.py still prints a deprecation warning saying you shouldn't call it but it still works.

pyproject.toml
bsteffensmeier: I forgot to remove that when I got the dynamic version working. Leaving the version specifier in the setup() call in setup.py allows it to pick up the version in version.py so we don't have to remember to keep this version in sync. I will remove the commented out version.

pyproject.toml
bsteffensmeier: I copied this from https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#declaring-the-build-backend.

ndjensen (approved):

pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+build-backend = "setuptools.build_meta:__legacy__"
+
+[project]
+name = "jep"
+description = "Jep embeds CPython in Java through JNI."
+dynamic = ["version"]
+license = { "text" = "zlib/libpng"}
```
bsteffensmeier, line 9: license is an attribute that confused me and may need more work in the future. We have the option to specify a license file or license text but when I tried the file it embedded the full license into the setup.cfg which is included in the upload to pypi and I am not sure how that shows up on pypi so I reverted to the simple text here which makes the setup.cfg match what it did before. 

It also seems that [PEP-639](https://peps.python.org/pep-0639/) was just approved and applied to the [pyproject.toml specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/) last month. According to the current documentation the license needs to be a valid [SPDX identifier](https://spdx.org/licenses/) which this doesn't seem to be. But the versions of pip I was testing with don't support PEP-639 so I don't think we can actually update until we drop support for all older versions of Python.

ndjensen: I think ours is the one listed as zlib.  https://opensource.org/license/zlib

bsteffensmeier: Do you think I should change this to zlib?

ndjensen: Probably should be Zlib to match the SPDX identifier?  The SPDX page doesn't mention libpng but the OSI page does, so I think it's the same thing.

Issue #593, opened 2025-01-23
Independent SubInterpreters are still not concurrent with python 3.12+


novos40:
**Describe the bug**
I'm running latest python 3.13.1 on 8 core VM. I'm starting 8 java threads. Each thread creates it's own SubInterpreter and runs CPU-only function like
```
def cpu_bound(number):
    # print(f">>> cpu_bound({number = })")                   # debug print
    # res = sum(i * i for i in range(number))                # CPU load
    print(f">>> cpu_bound({number = }): while loop")
    res = 0
    i = 0
    while(i < number):
        res += i * i
        i += 1
    print(f"<<< cpu_bound({number = }): while loop")
    return res
```
I've removed any functions call even functions like sum(). It seems like SubInterpreter creations and even function starts are concurrent, but overall it's still synchronized on something (despite the fact that python 3.12+ should support per-interpreter GIL). The output looks like this (threads started with consequent numbers so we can match start and stop)
```
>>> cpu_bound(number = 100000001): while loop
>>> cpu_bound(number = 100000003): while loop
>>> cpu_bound(number = 100000004): while loop
>>> cpu_bound(number = 100000005): while loop
>>> cpu_bound(number = 100000007): while loop
>>> cpu_bound(number = 100000002): while loop
>>> cpu_bound(number = 100000006): while loop
>>> cpu_bound(number = 100000000): while loop
<<< cpu_bound(number = 100000003): while loop
<<< cpu_bound(number = 100000001): while loop
<<< cpu_bound(number = 100000007): while loop
<<< cpu_bound(number = 100000005): while loop
<<< cpu_bound(number = 100000004): while loop
<<< cpu_bound(number = 100000006): while loop
<<< cpu_bound(number = 100000002): while loop
<<< cpu_bound(number = 100000000): while loop
```
which tells me that all functions start concurrently, but then they wait for each other in some random order. In multiple runs the order of starts and stops can vary, but first thread finish is always only after last thread start. The CPU utilization is never exceed 12% (except very short start period for code compilation, I guess) i.e. exactly like single thread execution. For that matter you can start python threads and get exactly the same performance or rather lack thereof.

What am doing wrong here? How do I make sub-interpreters execute concurrently?

**To Reproduce**
1. Create and start java platform thread
2. Create independent SubInterpreter in each thread
3. Call python function

**Expected behavior**
Fully independent SubInterpreters should run concurrently allowing for 100% CPU utilization.

**Environment (please complete the following information):**
[...]

PR #594, opened 2025-02-12
Make all jep types compatible with isolated interpreters.

bsteffensmeier:
This change makes it so that all the jep types declared in c code are compatible with sub interpreters. PyJObject and its subclasses(pyjclass and pyjarray) are converted to heap types and each interpreter now has a separate instance of those types that is stored in the module state for the _jep module. All other static types created by jep are now flagged as immutable which allows them to be shared between isolated interpreters.

In my tests this change resolves the crash seen in #593.

### DIFF
--- a/src/main/c/Objects/pyjarray.c
+++ b/src/main/c/Objects/pyjarray.c
@@ -37,24 +37,21 @@ static int pyjarray_init(JNIEnv*, PyJArrayObject*, int, PyObject*);
 static Py_ssize_t pyjarray_length(PyObject *self);
 
 
-
 // called internally to make new PyJArrayObject instances
 PyObject* pyjarray_new(JNIEnv *env, jobjectArray obj)
 {
     PyJArrayObject  *pyarray;
-
-    if (!PyJArray_Type.tp_base) {
-        PyJArray_Type.tp_base = &PyJObject_Type;
-    }
-    if (PyType_Ready(&PyJArray_Type) < 0) {
-        return NULL;
-    }
     if (!obj) {
         PyErr_Format(PyExc_RuntimeError, "Invalid array object.");
         return NULL;
     }
 
-    pyarray = (PyJArrayObject*) PyJObject_New(env, &PyJArray_Type, obj, NULL);
+    JepModuleState* modState = pyembed_get_module_state();
+    if (!modState) {
+        return NULL;
+    }
+    pyarray = (PyJArrayObject*) PyJObject_New(env, modState->PyJArray_Type, obj,
+              NULL);
     if (!pyarray) {
         return NULL;
     }
@@ -88,13 +85,6 @@ PyObject* pyjarray_new_v(PyObject *isnull, PyObject *args)
     // args
     PyObject *one, *two, *three;
     one = two = three = NULL;
-
-    if (!PyJArray_Type.tp_base) {
-        PyJArray_Type.tp_base = &PyJObject_Type;
-    }
-    if (PyType_Ready(&PyJArray_Type) < 0) {
-        return NULL;
-    }
 
     env = pyembed_get_env();
 
@@ -262,7 +252,12 @@ PyObject* pyjarray_new_v(PyObject *isnull, PyObject *args)
         return NULL;
     }
 
-    pyarray = (PyJArrayObject*) PyJObject_New(env, &PyJArray_Type, arrayObj, NULL);
+    JepModuleState* modState = pyembed_get_module_state();
+    if (!modState) {
+        return NULL;
+    }
+    pyarray = (PyJArrayObject*) PyJObject_New(env, modState->PyJArray_Type,
+              arrayObj, NULL);
     if (!pyarray) {
         return NULL;
     }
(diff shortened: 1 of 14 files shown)


src/main/c/Objects/pyjclass.c
```diff
@@ -449,6 +462,8 @@ PyMethodDef pyjclass_methods[] = {
 
 
 static PyMemberDef pyjclass_members[] = {
+    /* TODO Consider moving to managed dict as described in the documentation for tp_dictoffset */
+    {"__dictoffset__", T_PYSSIZET, offsetof(PyJClassObject, attr), READONLY},
```
ndjensen, line 466: What does this change do?

src/main/c/Objects/pyjarray.c
```diff
@@ -1626,69 +1624,29 @@ PyMethodDef pyjarray_methods[] = {
...
-    (lenfunc) pyjarray_length,                /* mp_length */
-    (binaryfunc) pyjarray_subscript,          /* mp_subscript */
-    0,                                        /* mp_ass_subscript */
-};
-
-
 static PyObject* pyjarray_iter(PyObject *);
 
+static PyType_Slot slots[] = {
+    {Py_tp_doc, (void*) &list_doc},
```
ndjensen, line 1630: I'm not sure I agree with using list_doc, though I'm not entirely sure what this is used for.  If you open the jep interpreter and make a pyjarray and do .__doc__ on it, what does it say?

src/main/c/Objects/pyjclass.c
```diff
@@ -458,43 +473,21 @@ static PyGetSetDef pyjclass_getset[] = {
...
+    {Py_tp_getattro, pyjclass_getattro},
+    {Py_tp_setattro, pyjclass_setattro},
+    {Py_tp_methods, pyjclass_methods},
+    {Py_tp_members, pyjclass_members},
+    {Py_tp_getset, pyjclass_getset},
+    {0, NULL},
+};
+
+PyType_Spec PyJClass_Spec = {
+    .name = "java.lang.Class",
```
ndjensen, line 489: I take it this was intentional that you changed the name?  Will that affect anything?  I noticed below that PyJObject has java.lang.Object so this change makes it more consistent at least.

src/main/c/Objects/pyjmultimethod.c
```diff
@@ -239,7 +239,8 @@ PyTypeObject PyJMultiMethod_Type = {
     0,                                        /* tp_getattro */
     0,                                        /* tp_setattro */
     0,                                        /* tp_as_buffer */
-    Py_TPFLAGS_DEFAULT,                       /* tp_flags */
+    Py_TPFLAGS_DEFAULT |
+     Py_TPFLAGS_IMMUTABLETYPE,                /* tp_flags */
```
ndjensen, line 243: Got an extra space in here.

src/main/c/Objects/pyjtype.c
```diff
@@ -49,7 +43,7 @@ static int staticTypesInitialized = 0;
 static PyTypeObject* addCustomTypeToTypeDict(JNIEnv *env, PyObject* fqnToPyType,
         jclass class, PyTypeObject *type)
 {
-    if (PyDict_SetItemString(fqnToPyType, type->tp_name, (PyObject*) type)) {
+    if (PyDict_SetItemString(fqnToPyType, type->tp_name, (PyObject * ) type)) {
```
ndjensen, line 46: Is there a reason there are spaces around the * now?  You did it multiple times in this file.

src/main/c/Objects/pyjclass.c
bsteffensmeier: This is the equivalent of [tp_dictoffset](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dictoffset) which was used previously. The documentation for [PyType_Slot](https://docs.python.org/3/c-api/type.html#c.PyType_Slot.slot) mentions that no slot is provided for tp_dictoffset and recommends switching to a managed dict, but I couldn't get that working I so followed the fallback recommendation of putting the offset in the members.

src/main/c/Objects/pyjarray.c
bsteffensmeier: This is the only place list_doc is used and I am pretty sure it is the same behavior as before. pyjarray.doc is not a valid attribute but __doc__ is a string with list_doc in it. The main purpose of this is so that at an interactive interpreter help(jep.jarray) shows a nice help message.

src/main/c/Objects/pyjclass.c
bsteffensmeier: It was not intentional, thanks for pointing it out. I am actually opposed to naming this java.lang.Class because class methods like getSuperclass() don't work on PyJClass so I think I should revert the name back to jep.PyJClass.

src/main/c/Objects/pyjtype.c
bsteffensmeier: astyle automatically added that in. I am using the same astyle configuration I have been using for years that I think I originally got from you but it is a newer version of astyle so it might have a slightly different default. I like the old way better so I will look into it and see if that is something I can change in the astyle configuration.

bsteffensmeier: This seems to be a bug with astyle 3.4.13, I've also tried 3.6.6 and I see the same behavior. Reverting to astyle 3.3 produces better results so I am going with that for this change.

src/main/c/Objects/pyjmultimethod.c
bsteffensmeier: Fixed

ndjensen (approved):

PR #596, opened 2025-02-19
Expose PyConfig options for isolation and arg parsing.

bsteffensmeier:
This builds on the changes in #586 and further improves the compatibility with the c-api. In that change I hardcoded the option to disable argv parsing however there may be use cases where someone wants to enable that parsing. To keep things consistent with the c-api the option is exposed in Java and Jep will no longer set the value of that option unless it is set in Java.

This also exposes the ability to create an isolated python configuration. For backwards compatibility the default is to use a config which behaves like regular python but the c-api includes a function to create an isolated python so I thought it would be useful to allow that option in Java.

### DIFF
--- a/src/main/c/Jep/pyembed.c
+++ b/src/main/c/Jep/pyembed.c
@@ -287,11 +287,13 @@ static void handle_startup_exception(JNIEnv *env, const char* excMsg)
 
 
 void pyembed_startup(JNIEnv *env,
+                     jboolean isolated,
                      jobjectArray argv,
                      jint hashSeed,
                      jint useHashSeed,
                      jstring home,
                      jint optimizationLevel,
+                     jint parseArgv,
                      jstring programName,
                      jint siteImport,
                      jint useEnvironment,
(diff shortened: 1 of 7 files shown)


src/main/c/Jep/pyembed.c
```diff
@@ -356,9 +358,15 @@ void pyembed_startup(JNIEnv *env,
...
+    if (isolated) {
+        PyConfig_InitIsolatedConfig(&config);
+    } else {
+        PyConfig_InitPythonConfig(&config);
+    }
     // According to PEP-587 the fields shared with PyPreConfig should be set first.
-    config.parse_argv = 0;
+    if (parseArgv >= 0) {
+        config.parse_argv = 0;
+    }
```
ndjensen, line 369: This confuses me.  The options are 0 or 1, so how do you set it to 1 with this code?    https://docs.python.org/3/c-api/init_config.html#c.PyConfig.parse_argv

bsteffensmeier: That is definitely an oversight on my part. Thanks for pointing it out. I've fixed it in the latest commit.

PR #600, opened 2025-03-13
Fix cases where startup errors can cause crashes.

bsteffensmeier:
Using invalid SubInterpreterOptions was causing a crash rather than throwing an exception even though we have checks to throw an exception. It turns out we were trying to throw a JepException before initializing the variable that holds JepException is initialized. We have an existing function to handle startup errors before JepException is available but that function checks for python exceptions and errors during python initialization probably shouldn't use the python APIs. This change makes the python error check optional in `handle_startup_exception` and changes any startup errors that use `THROW_JEP` to use `handle_startup_exception` instead.

### DIFF
--- a/src/main/c/Jep/pyembed.c
+++ b/src/main/c/Jep/pyembed.c
@@ -216,7 +216,7 @@ static int initjep(JNIEnv *env, jboolean hasSharedModules)
     PyObject *sysmodules = PyImport_GetModuleDict();
     PyObject *modjep = PyDict_GetItemString(sysmodules, "_jep");
     if (!modjep && PyErr_Occurred()) {
-        handle_startup_exception(env, "Error checking for exisitng module _jep", 1);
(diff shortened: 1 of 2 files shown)


src/main/c/Jep/pyembed.c
```diff
@@ -216,16 +216,16 @@ static int initjep(JNIEnv *env, jboolean hasSharedModules)
     PyObject *sysmodules = PyImport_GetModuleDict();
     PyObject *modjep = PyDict_GetItemString(sysmodules, "_jep");
     if (!modjep && PyErr_Occurred()) {
-        handle_startup_exception(env, "Error checking for exisitng module _jep");
+        handle_startup_exception(env, "Error checking for exisitng module _jep", 1);
```
ndjensen, line 219: Existing problem but should be fixed, the spelling of "existing".

ndjensen (approved):
Good improvements!

Issue #599, opened 2025-03-12
Fatal error when running multiple SubInterpreters in parallel with OWN_GIL=true


matias-santurio:
**Describe the bug**
A fatal JVM error occurs I try to run multiple SubInterpreters in parallel (each in its own thread) when OWN_GIL is set to true.
The error trace I see is: 

```
 [thread 219 also had an error][thread 218 also had an error]
 [thread 214 also had an error]
 
 [thread 212 also had an error]
 #
 # A fatal error has been detected by the Java Runtime Environment:
 #
 #  SIGSEGV (0xb)[thread 220 also had an error]
  at pc=0x00000041a4cc14c5, pid=64, tid=222
 #
 # JRE version: OpenJDK Runtime Environment Corretto-21.0.5.11.1 (21.0.5+11) (build 21.0.5+11-LTS)
 # Java VM: OpenJDK 64-Bit Server VM Corretto-21.0.5.11.1 (21.0.5+11-LTS, mixed mode, emulated-client, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, linux-amd64)
 # Problematic frame:
 # C  [libpython3.12.so.1.0+0x1db4c5][thread 215 also had an error]
 [thread 216 also had an error]
 [thread 213 also had an error]
   _PyObject_Malloc+0x75
 #
 # Core dump will be written. Default location: /app/core
 #
 [thread 221 also had an error]
 # An error report file with more information is saved as:
 # /app/hs_err_pid64.log
 [thread 211 also had an error]
 [thread 217 also had an error]
 [12.100s][warning][os] Loading hsdis library failed
 #
 # If you would like to submit a bug report, please visit:
 #   https://github.com/corretto/corretto-21/issues/
 # The crash happened outside the Java Virtual Machine in native code.
 # See problematic frame for where to report the bug.
 #
```

**To Reproduce**
The error can be consistently be reproduced from the following code snippet in a `linux/amd64` architecture. Couldn't reproduce on `linux/aarch64`. 

```java

import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.CompletableFuture;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import jep.Interpreter;
import jep.JepConfig;
import jep.MainInterpreter;
import jep.SubInterpreter;
import jep.SubInterpreterOptions;

public class Application {
  private static final String JEP_PATH_ENV_VAR = "JEP_PATH";

  /**
[...]

PR #601, opened 2025-03-16
Convert all types to heap types.

bsteffensmeier:
The primary goal of this change is to fix the crashes described in #599 when isolated sub-interpreters are used. The crashes are mostly a problem in the case where all interpreters are initialized on different threads at the same time.

I was not able to track down the exact source of crashes in the cpython code but it seems to be related to the internal type cache. I found that the crashes would not occur if I stopped using all our static types which also required removing a bunch of jep functionality. Since heap types are specific to an interpreter they are safer for isolated interpreters so I started testing using more heap types instead of removing functionality. Using heap types stopped the crashing. This change converts all of the jep types that were previously static types to be heap types instead. I have not proven definitively that all the types were involved in the crashes but a majority are so it seemed beneficial to do all of them for consistency.

The biggest downside to heap types is the type object can no longer be referenced directly in c. I extended the approach from #594 to store the type definitions in the module state. In the code that creates new PyJTypes there is a need to create many super types, methods, and fields and that needs a reference to the module state. To avoid redundant lookups of the module state when making new PyJTypes the module state is passed around between most of the pyjtype functions.

Another benefit of using heap types is that PyTypeObject used by static types is not part of the cpython Limited ABI but the functions for initializing the types are. This change brings us a little closer to being able to use the stable ABI to distribute a single python binary for multiple versions of cpython.

### DIFF
--- a/src/main/c/Objects/pyjfield.c
+++ b/src/main/c/Objects/pyjfield.c
@@ -36,6 +36,7 @@
 static void pyjfield_dealloc(PyJFieldObject *self)
 {
 #if USE_DEALLOC
+    PyTypeObject *tp = Py_TYPE(self);
     JNIEnv *env  = pyembed_get_env();
     if (env) {
         if (self->rfield) {
@@ -46,6 +47,7 @@ static void pyjfield_dealloc(PyJFieldObject *self)
     Py_CLEAR(self->pyFieldName);
 
     PyObject_Del(self);
+    Py_DECREF(tp);
 #endif
 }
 
--- a/src/main/c/Objects/pyjfield.c
+++ b/src/main/c/Objects/pyjfield.c
@@ -46,9 +46,17 @@ static void pyjfield_dealloc(PyJFieldObject *self)
 
     Py_CLEAR(self->pyFieldName);
 
-    PyObject_Del(self);
+    tp->tp_free((PyObject*) self);
     Py_DECREF(tp);
 #endif
+}
+
+
+static int pyjfield_traverse(PyJFieldObject *self, visitproc visit, void *arg)
+{
+    Py_VISIT(self->pyFieldName);
+    Py_VISIT(Py_TYPE(self));
+    return 0;
 }
 
 
@@ -58,7 +66,8 @@ PyJFieldObject* PyJField_New(JNIEnv *env, JepModuleState *modState,
     PyJFieldObject *pyf;
     jstring          jstr        = NULL;
 
-    pyf              = PyObject_NEW(PyJFieldObject, modState->PyJField_Type);
+    PyTypeObject* tp = modState->PyJField_Type;
+    pyf              = (PyJFieldObject*) tp->tp_alloc(tp, 0);
     pyf->rfield      = (*env)->NewGlobalRef(env, rfield);
     pyf->pyFieldName = NULL;
     pyf->fieldTypeId = -1;
@@ -645,6 +654,7 @@ static int pyjfield_descr_set(PyObject *field, PyObject *obj, PyObject *value)
 static PyType_Slot slots[] = {
     {Py_tp_doc, "jfield"},
     {Py_tp_dealloc, pyjfield_dealloc},
+    {Py_tp_traverse, pyjfield_traverse},
     {Py_tp_descr_set, pyjfield_descr_set},
     {Py_tp_descr_get, pyjfield_descr_get},
     {0, NULL},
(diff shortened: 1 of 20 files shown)


ndjensen (commented):
One minor comment.

ndjensen:
Also Appveyor does not like your new test.

src/main/c/Objects/pyjobject.c
```diff
@@ -77,8 +78,8 @@ static void pyjobject_dealloc(PyJObject *self)
             (*env)->DeleteGlobalRef(env, self->clazz);
         }
     }
-
-    Py_TYPE((PyObject*) self)->tp_free((PyObject*) self);
+    tp->tp_free((PyObject*) self);
```
ndjensen, line 81: Why do we call tp_free() on this one?

ndjensen (approved):
Good job finding and correcting the memory leak of heap types.

src/main/c/Objects/pyjobject.c
bsteffensmeier: PyJObject and the subtypes that we create manually like PyJArray and PyJClass do not use garbage collection because they can't be involved in reference cycles so they need to be freed with `PyObject_Del`. When we dynamically make subtypes for subclasses, like java.util.ArrayList or any other Java class then python automatically adds cgarbage collection so `PyObject_GC_Del` needs to be called to free the memory. `tp_free` is always set to the correct option from those two so we don't have to try anything else to figure out which one is right.

bsteffensmeier:
Although the initial change to heap types was relatively straightforward further testing showed that memory management is quite a bit more complicated with heap types so there are follow-on commits to address issues I found. I now believe this change is complete and properly manages all the memory associated with the module state and heap types.

To test for memory leaks I used a loop that endlessly creates and closes SubInterpreters and monitored the process memory to see if it stabilized. At first I was using Python 3.13 and I noticed substantial leaks but after going back and testing old versions of jep that should not be leaking I found the exact same memory leaks. I eventually found [a python bug](https://github.com/python/cpython/issues/110411) which indicates recent versions of python leak memory when sub-interpreters are closed so I used python 3.11 for testing for memory leak and found there was memory leaks related to this changes.

I tracked the memory leaks down to the new type objects and by tracking all ref count changes to the types I found the refcount increases when an instance is created but does not decrease when the instance is freed. In 7603147 I added a decref to the `tp_dealloc` for all the heap allocated types to address this problem.

After that change I was still seeing a memory leak but I could make it go away by removing the java import hook before closing an interpreter. I didn't like that as a solution but this told me that the reference counts must be correct since cpython can free all the memory as long as I let it free the java import hook before closing the interpreter. Eventually I solved that leak in 99889f5 by implementing m_traverse on the module to let GC know about cycles involving the _jep module. I was not able to track down exactly why m_traverse fixes it but I suspect there was a reference cycle involving the module and the system module which holds the java import hook.

After 99889f5 I have not seen any memory leak in normal use cases. However in tracking down the need to implement m_traverse I had also found [this python bug](https://github.com/python/cpython/issues/87138) and was beginning to suspect we may need to implement GC and tp_traverse for all our types. I was able to trigger a leak by adding a list as a member of the java import hook and then adding the list to itself to make a reference cycle and then putting a pyjobject in the list. It's a rather contrived test but it proves that without implementing GC there will always be ways to theroetically have enough reference cycles to trigger a leak when closing a sub-interpreter so I decided to go ahead and implement gc in all our types in 3d64318 which solves the leaks even in my contrived test.

Finally while writing this comment I was reviewing the [documentation for tp_dealloc](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dealloc) and for [Changing Static Types to Heap Types](https://docs.python.org/3/howto/isolating-extensions.html#changing-static-types-to-heap-types) and they both mention calling `PyObject_GC_UnTrack`. I was not seeing any problems with this missing call, I suspect we may not need it since our types aren't directly involved in any reference cycles, however I want to follow best practices so in 344d76e I added calls to `PyObject_GC_UnTrack`.

src/main/c/Objects/pyjfield.c
```diff
@@ -36,6 +36,9 @@
 static void pyjfield_dealloc(PyJFieldObject *self)
 {
 #if USE_DEALLOC
+    PyObject_GC_UnTrack(self);
+    PyObject_GC_UnTrack(self);
```
ndjensen, line 40: This line is in here twice.

ndjensen (approved):

PR #603, opened 2025-03-28
Do not trigger numpy import when converting non-numpy python objects to java

bsteffensmeier:
This change prevents jep from importing numpy if it isn't already imported into an interpreter. It is especially important for anyone who builds jep with numpy support and wants to use isolated sub-interpreters. Numpy does not work in isolated sub-interpreters so jep should not ever import it. It should still be possible to use the same build of Jep with numpy support in shared interpreters.

Without this change TestIsolatedSubInterpreter.java will fail when jep is built with numpy support so new testing is not needed, our existing test verify this change whenever jep is built with numpy support.

### DIFF
--- a/src/main/c/Jep/jep_numpy.c
+++ b/src/main/c/Jep/jep_numpy.c
@@ -73,10 +73,39 @@ static int init_numpy(void)
     return numpyInitialized;
 }
 
+/*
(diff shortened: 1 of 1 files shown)


ndjensen (approved):
A good change.

PR #616, opened 2025-09-22
Change maven groupId to org.ninia.

bsteffensmeier:


ndjensen (approved):
I really like this change!

Issue #633, opened 2026-05-03
Expose formatted Python traceback string on `JepException`


magicmq:
I propose adding a `getPythonTraceback()` method to `JepException`, which returns the Python traceback (String), formatted by the standard `traceback` module (I.E., what `traceback.print_exc()` would print). I can submit a pull request to implement this feature, but I wanted to check beforehand whether this is something you would be open to merging.

When an uncaught Python exception is converted to a `JepException`, JEP does a good job of synthesizing `StackTraceElement`s from the Python traceback so the combined Python and Java stack appears in the stack trace. It's useful for tools that walk `getStackTrace()` programmatically, but for applications that *log* errors, the synthesized Java-style stack loses a few things I would expect from a Python traceback, such as:

- The source line text (e.g. `test.test()`) under each frame
- PEP 657 carets (`^^^^`) pointing at the offending expression (on Python 3.11+)
- The familiar `Traceback (most recent call last):` / `ExceptionType: message` formatting
- Expected ordering of frames: in the Java stack trace, most recent calls are ordered first, but in a Python traceback, the most recent calls are last.

I also noticed that `SyntaxError`s don't reliably print the line number in the Java stack trace, which could make debugging more difficult.

This matters most for projects that embed JEP as a scripting layer for end users (my use case), where people reading error logs are Python script authors, not Java developers, and the Python-formatted traceback is what they would expect to see and get the most information from.

My current workaround is a Python-sided utility that catches `BaseException`, calls `traceback.format_exc()`, and stashes the result of that call somewhere Java can read it, before re-raising the exception:
```py
import traceback as __jep_tb

def __jep_trampoline(fn, *args, **kwargs):
    try:
        return fn(*args, **kwargs)
    except BaseException:
        ExceptionHandler.captureTraceback(__jep_tb.format_exc())
        raise
```

This works, but it adds a Python frame to every call across the JNI boundary. That isn't a whole lot of additional overhead, but the solution seems somewhat tacky to me, especially when I can call directly from Java into Python with `PyCallable.callAs(...)`. It seems much more intuitive to just cache the formatted traceback in `JepException`, and then fetch the traceback by catching `JepException` on the Java side.

The `getPythonTraceback()` method would return a formatted Python traceback (or `null` if no Python traceback is available, either because the exception did not originate in Python, or an error occurred when formatting the traceback). I believe this would only require minor modifications to `JepException.java` and `jep_exceptions.c`. I am happy to put together a PR with tests.

[...]

Issue #319, opened 2021-03-17
Pass JepException back into jep


jsnps:
Hey, 

Is it possible to pass some caught JepException back into python to handle the original python exception in python code?
I've read that Jep is able to throw back some JepException into python, I guess this is the case for reentrant calls (e.g. java -> python -> java -> python). 
Background is, that I want to print the python stack trace when runnning a script and something goes wrong. 
So something like this:

```java
		try (Jep jep = new SharedInterpreter()) {
			try {
				jep.runScript("faultyScript.py");
			} catch (JepException e) {
				//how to get the exception back into python?
				jep.set("ex", e);
				jep.exec("import traceback");
				jep.exec("traceback.print_exception(ex, ex.args, ex.__traceback__)");
			}
		}
```
The issue here is that the exception is a PyObject, a simple python wrapper around the java exception object.

So far I only see some alternative like this:

```java
		try (Jep jep = new SharedInterpreter()) {
			try {
				jep.eval("try:\n" +
					"  exec(open(\"faultyScript.py\").read())\n" + 
					"except Exception as e:\n" + 
					"  import traceback\n" + 
					"  traceback.print_exc()\n" + 
					"  raise e\n");
			} catch (JepException e) {
				//more handling on java side
			}
		}
```

Although, for this approach I would need to massage the stacktrace a bit. Any other ideas?

sys.last_value only seems to be set in an interactive interpreter, sys.exc_info() also doesn't seem to be set, which is kind of excepted, because the exception is caught on the java side. 

Second alternative, add a showException flag to runScript and do this in C?

bsteffensmeier:
There is currently no way to get the python exception from a JepException. I have considered adding the python exception to the JepException as a PyObject but it never seemed like that would be particularly useful and it could cause some confusion for cases where the JepException is thrown outside the scope of the Interpreter which created the PyObject. If you have a need for that info though it is something I would consider merging if you can make a pull request.

You may be able to create your own java Exception class that contains a PyObject. I don't think you could throw that exception directly form python but you could have a java function throw it. Your exception containing the PyObject would then be the cause of the JepException. See the code below for an example. The traceback wasn't working but the idea of storing the exception as a PyObject seems workable. You may need a separate field in the PythonExceptionWrapper for the traceback to make sure it gets across. You could store any state information you want in the PythonExceptionWrapper for use later in python.


```java
class Test {

    public static class PythonExceptionWrapper extends RuntimeException {
        public final PyObject cause;

[...]

PR #634, opened 2026-05-06
Attach PyObject to exceptions, fix memory leak when constructing Java `StackTraceElement`s

magicmq:
## Changes

The below changes are non-breaking.

### Minor changes to exception handling

This PR changes exception handling within an interpreter context, for the following scenarios:

- If a *pure* Python exception is raised from Python (such as `ValueError`, `SyntaxError`, `ArithmeticError`, etc.), a `PythonException` object is instantiated and thrown (as long as the exception isn't caught in Python). The underlying Python exception object is attached to the exception (as a `PyObject`) and is obtainable via `PythonException.getPythonException()`. For example:

```java
JepConfig config = new JepConfig().addIncludePaths(".");
try (Interpreter interp = new SubInterpreter(config)) {
    try {
        interp.eval("raise ValueError('Test traceback')");
    } catch (PythonException e) {
        PyObject pyExc = e.getPythonException();
        // Do something with the pyExc...
    }
}
```

- If a Java exception is surfaced as a result of a Python call, a `JepException` object is instantiated and thrown. For example:

```java
JepConfig config = new JepConfig().addIncludePaths(".");
try (Interpreter interp = new SubInterpreter(config)) {
    interp.eval("from java.util import ArrayList");
    try {
        interp.eval("ArrayList().get(0)");
    } catch (PythonException e) {
        // This will NOT be caught
    } catch (JepException e) {
        // This WILL be caught
        Throwable cause = e.getCause();
        // cause will be an instance of IndexOutOfBoundsException
    }
}
```

The following changes were also made:
- A `getPythonTraceback()` method was added to `JepException`: a traceback may be available for both of the above scenarios, while a Python exception object will only be available for `PythonException`.
- A `process_python_traceback` function was added to `jep_exceptions.c` to provide the formatted Python traceback String and the frames for subsequent processing into Java `StackTraceElement`s.
- The `TestExceptionCause` test was merged into a new test, `TestJepException`, which performs more granular testing on `JepException` (and `PythonException`) in a variety of possible scenarios.

Commit 4e547e979196a0373f914cc25a4c4a71430f1707 makes the above changes/additions.

#### Quick Aside

I think the above changes could potentially create confusion, particularly with the second scenario: users may expect all exceptions that "came from" Python code to be a `PythonException`, even if the exception originated in Java. I can't think of a better way to solve this without making breaking changes to the existing code, but nevertheless, I attempted to make this as clear as possible in the JavaDocs.

Although this PR solves the issues outlined in #319 and #633, I think it would make a lot more organizational sense to do the following:

- Reserve `JepException` for JEP-related exceptions that occur *outside* of the scope of an interpreter (for example, creating a `JepConfig` after creating a `SharedInterpreter`, or throwing to signal invalid thread access).
- Create a new class called `InterpreterException`, which becomes a new superclass for any exceptions thrown *inside* the interpreter context (either pure Python exceptions or Java exceptions which surface as a result of Python calls). I think it would be useful to attach a formatted Python traceback String to this superclass. 
- Make `PythonException` a subclass of `InterpreterException`, and throw this subtype for *pure* Python exceptions only. It would cache the underlying Python exception as a `PyObject`.
- Create `JavaException`, a subclass of `InterpreterException`, and throw this subtype for Java exceptions that occur as a result of a Python call. The underlying Java exception would be obtainable via `getCause()`, as before. Although the cause was a Java exception, it may still be useful to have a Python traceback, so that's obtainable via the superclass' `InterpreterException.getPythonTraceback()` method.

I didn't do this because it would introduce breaking changes, since `JepException` currently encompasses exceptions thrown in a Python context.

### Patch for memory leak when handling exceptions

This PR also fixes a memory leak in `jep_exceptions.c`. When looping over the Python traceback to build Java `StackTraceElement`s, `PySequence_GetItem` creates a new reference, but these new references are immediately passed to other functions, without decrefing:

```c
stackEntry = PyList_GetItem(pystack, i);
// java order is classname, method name, filename, line number
// python order is filename, line number, function name, line
charPyFile = PyUnicode_AsUTF8(
                 PySequence_GetItem(stackEntry, 0));
pyLineNum = (int) PyLong_AsLongLong(
                PySequence_GetItem(stackEntry, 1));
charPyFunc = PyUnicode_AsUTF8(
                 PySequence_GetItem(stackEntry, 2));
pyLine = PySequence_GetItem(stackEntry, 3);
```

In addition, `pyLine` is never decrefed. This leaks up to 4 Python objects per stack frame.

This is fixed by saving all four references as named variables, and explicitly calling decref on them when apporpriate.

Commit 6ea92171628bf4cf23ab0405b3c47c8260bb42e2 fixes this memory leak.


bsteffensmeier:
> I think the above changes could potentially create confusion, particularly with the second scenario: users may expect all exceptions that "came from" Python code to be a `PythonException`, even if the exception originated in Java. I can't think of a better way to solve this without making breaking changes to the existing code, but nevertheless, I attempted to make this as clear as possible in the JavaDocs.

I have started looking at this and overall it looks good so far. I will need more time to fully review it but for now I keep coming back to this point. Your documentation and explanation here make me feel like I could comfortably use the API correctly but I don't understand why we need to structure it that way. Is there a technical or compatibility reason why a Python Exception that was originally a Java Exception can't be a PythonException instead of a JepException? I think we have a python exception object and it might not be as interesting as the original java cause but it should still technically be enough to satisfy the requirements to make a PythonException.

> ### Patch for memory leak when handling exceptions

Awesome!

magicmq:
> Is there a technical or compatibility reason why a Python Exception that was originally a Java Exception can't be a PythonException instead of a JepException? I think we have a python exception object and it might not be as interesting as the original java cause but it should still technically be enough to satisfy the requirements to make a PythonException.

I didn't necessarily make it that way because of a technical or compatibility issue, but a Python exception that was originally a Java exception wouldn't have a Python exception object, would it? I know JEP converts a few select Java exceptions into equivalent Python exceptions (and in those cases, Python exception objects would be available), but my understanding was that this was done solely to facilitate catching those Java exceptions in Python, not necessarily because the Python exception object adds any value or additional information to the exception itself.

I suppose that a `PythonException` could be thrown as well for Java exceptions that surface from Python, but that seemed even more misleading than the current design, since the exception wasn't really a Python exception at all, it was a Java exception that only happened to surface as a result of a Python call.

bsteffensmeier:
> I didn't necessarily make it that way because of a technical or compatibility issue, but a Python exception that was originally a Java exception wouldn't have a Python exception object, would it?

In `process_py_exception()` there is always a python exception object, the only time that wasn't the case is when we weren't calling `PyErr_NormalizeException()`, but with that addition we will always have a Python exception, whether the original cause was Python related or from Java.

In fact, I just realized that with this new code, jpyException will always be a Java PyObject wrapping that python exception, even when the exception originated in java. For the case where you currently create a JepException we create jpyException without ever using it.

> I know JEP converts a few select Java exceptions into equivalent Python exceptions (and in those cases, Python exception objects would be available), but my understanding was that this was done solely to facilitate catching those Java exceptions in Python, not necessarily because the Python exception object adds any value or additional information to the exception itself.

Even in the case where there is no equivalent Python exception we create a Python RuntimeError so the lack of an equivalent exception type doesn't prevent the creation of a Python exception. I agree that the original intent was to make it easier to catch Java Exceptions in Python and there isn't much value in the Python Exception. But it seems to me this feature proves there is some value in the Python exception. The Python traceback is an attribute of a Python exception, you want access to this traceback even when an exception originated in Java so in this use case there is some value in the Python exception even if it originated from a Java exception.

You find the formatted traceback string to be useful but I can imagine there might be use cases where the traceback objects themselves might be useful for similar reasons. I see including the whole PyObject in the PythonException as a convenient way to just include all the context we have around the exception from Python in case anyone catching it wants to use it. I don't see any reason to ever include the formatted traceback and not give access to the whole exception.

> I suppose that a `PythonException` could be thrown as well for Java exceptions that surface from Python, but that seemed even more misleading than the current design, since the exception wasn't really a Python exception at all, it was a Java exception that only happened to surface as a result of a Python call.

This seems less confusing to me. If there was ever a Python exception then I think Jep should throw PythonException, even in the case where that PythonException is just triggered by a Java exception. I think this is similar to when a Java Exception has a long chain of CausedBy, often this is confusing and some of those layers tend to be meaningless but trying to cut out layers of causes from java exception is only going to make it more confusing.

ndjensen:
I don't understand the need to limit PythonException to pure Python exceptions.  If you call Java from Python, and that Java throws an exception, it will get turned into a Python exception so that the exception can be caught by Python code, if desired.  While the traceback may not be as useful in that case as the Java stacktrace, it still has some relevant information.  So I think most or all times the exception comes from executing Python code in the Python interpreter, even if it crosses over to Java, it should throw a PythonException.  I would think that we would want PythonException extends JepException and PythonException has the original exception PyObject in it in case someone wants to do something fancy with it.  The formatted traceback could be calculated on demand by using the PyObject, or if that's too cumbersome, we could create the formatted traceback when creating the PythonException.

I do think if there's ever a Jep 5 we should rework the exception hierarchy a bit, maybe making use of IllegalStateException more and then distinguishing between exceptions like config exceptions vs invalid thread exceptions vs the interpreter exceptions.

Thanks for finding the memory leak!

magicmq:
Ultimately, my choice to limit `PythonException` to pure Python exceptions was purely an organizational one- I didn't structure it that way out of necessity or because of technical/compatibility limitations. `PythonException` could very well encompass both native Python exceptions as well as Java exceptions that surface as a result of a Python call.

> In fact, I just realized that with this new code, jpyException will always be a Java PyObject wrapping that python exception, even when the exception originated in java. For the case where you currently create a JepException we create jpyException without ever using it.

In light of this, I'm in agreement that it makes much more sense to handle both of the above cases via `PythonException`, and just throw that for both. Previously, I was under the assumption that Java exceptions thrown as a result of Python activity _did not_ generate a `PyObject`, but I was mistaken.

Regarding your points made about having only the traceback versus the whole `PyObject`: The `PyObject` is available in `PythonException`; I didn't make any further changes to that. What do you guys think about making both the `PyObject` and the traceback string available in `PythonException`? For my specific use case, I'm only interested in the traceback; perhaps there are others who are in the same boat. The overhead of having to feed the `PyObject` back into Python only to extract the traceback out of it seems unnecessary, when the traceback could just be included alongside the `PyObject`. Granted, I haven't tested the performance cost of my `format_python_traceback` function in jep_exceptions.c, but I don't think it would be that expensive.

Regardless, I think I've addressed your guys' above concerns in my latest commit, do let me know your thoughts.

P.S. Sorry for leaving this unanswered for a while, life got in the way.

src/main/c/Jep/jep_exceptions.c
```diff
@@ -29,10 +29,118 @@
...
+                                              ptrace ? ptrace : Py_None,
+                                              NULL);
+        if (tbList != NULL) {
+            sep = PyUnicode_FromString("");
+            if (sep != NULL) {
+                joined = PyUnicode_Join(sep, tbList);
+                if (joined != NULL) {
+                    utf8 = PyUnicode_AsUTF8(joined);
+                    if (utf8 != NULL) {
+                        result = (*env)->NewStringUTF(env, utf8);
```
bsteffensmeier, line 80: Python is using standard utf-8 and Java is using a modified utf-8 so in theory there are strings where this won't work. We should consider using PyObject_to_jstring() which uses UTF-16 as the intermediary for better compatibility with characters beyond the UTF BMP.

We don't do this consistently for existing code so it's not something that must change to merge this PR. At some point I may go and try to fix all instances where we do this and if that ever happens this would be included but since no one ever complains about it I haven't made it a priority.

src/main/c/Jep/jep_exceptions.c
bsteffensmeier, line 249: Since we are now calling PyErr_NormalizeException I think this check and these comments are unnecessary.

bsteffensmeier (approved):
This looks good to me. My 2 minor comments are not blocking issues.

bsteffensmeier:
> What do you guys think about making both the `PyObject` and the traceback string available in `PythonException`?

That works for me, I think that having the traceback string available will let most users keep their exception handling code simpler which is generally better.

> P.S. Sorry for leaving this unanswered for a while, life got in the way.

No problem, this project isn't moving terribly fast. Along those lines, I think this change is a little more than I want to put in 4.3.2 so I plan to put it in 4.4. I haven't even made a branch for dev_4.4, I was planning to wait until after I release 4.3.2 and a couple small issues came in this week that might be going into 4.3.2. So I might not actually merge this PR for awhile but I am planning on merging it for Jep 4.4. I will release 4.4 after I have tested against the new Python 3.15 which is coming in October so that would be the earliest this would make it into a release

src/main/java/jep/PythonException.java
```diff
@@ -0,0 +1,119 @@
+/**
+ * Copyright (c) 2004-2022 JEP AUTHORS.
```
ndjensen, line 2: We're pretty inconsistent on this but it should probably be just 2026, not a range.  Not really worth another commit though.

src/main/c/Jep/jep_exceptions.c
```diff
@@ -100,6 +209,39 @@ int process_py_exception(JNIEnv *env)
...
+         * Capture pvalue as a Java PyObject now, before the block below may
+         * replace pvalue with args[0] (the string message). The Py_INCREF
+         * inside PyObject_As_JPyObject keeps the exception object alive
+         * independently of what happens to pvalue afterwards.
+         */
+        jobject jpyException = NULL;
+        if (pvalue != NULL) {
+            jpyException = PyObject_As_JPyObject(env, pvalue);
+            if ((*env)->ExceptionCheck(env)) {
+                (*env)->ExceptionClear(env);
```
ndjensen, line 240: Should there be an `(*env)->ExceptionDescribe(env);` before the exception clear?

ndjensen (approved):
This looks good to me.  I think it's fine to create the formatted traceback on every exception thrown.  I don't think people expect exception throwing code to be super performant.  I agree this should probably go in Jep 4.4.

PR #646, opened 2026-07-20
Guard NULL from GetStringUTFChars in raiseTypeError

### LINKED ISSUE #643: [BUG] GetStringUTFChars NULL unchecked in raiseTypeError — JVM crash on OOM
## Bug Description

In `raiseTypeError`, the return value of `GetStringUTFChars` is not checked for NULL before being passed to `PyErr_Format` as a `%s` argument. Per JNI Specification §6.2, `GetStringUTFChars` returns NULL when native memory allocation fails. Passing NULL to `%s` in `PyErr_Format` (which uses `printf`-style formatting) causes a SIGSEGV on most platforms.

## Location

`src/main/c/Jep/convert_p2j.c`, lines ~59-65

## Code

```c
expTypeName = (*env)->GetStringUTFChars(env, expTypeJavaName, 0);
// No NULL check — expTypeName could be NULL on OOM
if (PyJClass_Check(pyobject)) {
    actTypeName = "java.lang.Class";
} else {
    actTypeName = pyobject->ob_type->tp_name;
}
PyErr_Format(PyExc_TypeError, "Expected %s but received a %s.", expTypeName,
             actTypeName);
// expTypeName is NULL → printf %s dereferences NULL → SIGSEGV
(*env)->ReleaseStringUTFChars(env, expTypeJavaName, expTypeName);
```

## Context

`raiseTypeError` is called on every Python-to-Java type conversion mismatch. This is a routine error path — users passing the wrong type to a Java method will trigger it. The OOM that makes `GetStringUTFChars` return NULL is rare on its own, but under memory pressure (large JNI workloads, many threads), it becomes reachable.

Note that the function already has a `process_java_exception` check for the preceding `getName` call (line 56-58), so the pattern of checking JNI returns is established — this one was just missed.

## Impact

JVM crashes with SIGSEGV in the error-reporting path, which is particularly hard to debug since the crash happens while trying to report a different error (type mismatch).

## Suggested Fix

```c
expTypeName = (*env)->GetStringUTFChars(env, expTypeJavaName, 0);
if (expTypeName == NULL) {
    (*env)->DeleteLocalRef(env, expTypeJavaName);
    return;  // JVM has already thrown OutOfMemoryError
}
```

## Environment

- Jep: current master branch (commit 247887d)
- Affected: any Python-to-Java call where type mismatch occurs under memory pressure

SAY-5:
Fixes #643. In `raiseTypeError`, the return value of `GetStringUTFChars` was passed straight to `PyErr_Format` as a `%s` argument. Per the JNI spec that call returns NULL if it can't allocate the native string, and formatting NULL with `%s` crashes. This falls back to "<unknown>" for the message and skips the matching `ReleaseStringUTFChars` when nothing was returned, so the type-error path stays intact on that rare allocation failure.

### DIFF
--- a/src/main/c/Jep/convert_p2j.c
+++ b/src/main/c/Jep/convert_p2j.c
@@ -56,6 +56,14 @@ static void raiseTypeError(JNIEnv *env, PyObject *pyobject, jclass expectedType)
         return;
     }
     expTypeName = (*env)->GetStringUTFChars(env, expTypeJavaName, 0);
+    if (expTypeName == NULL) {
+        /*
+         * GetStringUTFChars can fail and leave a pending JNI exception (such as
+         * OutOfMemoryError). Clear it so it does not interfere with later JNI
+         * calls; the type error is still reported with a fallback name below.
+         */
+        (*env)->ExceptionClear(env);
+    }
     if (PyJClass_Check(pyobject)) {
         actTypeName = "java.lang.Class";
     } else {
--- a/src/main/c/Jep/convert_p2j.c
+++ b/src/main/c/Jep/convert_p2j.c
@@ -61,9 +61,11 @@ static void raiseTypeError(JNIEnv *env, PyObject *pyobject, jclass expectedType)
     } else {
         actTypeName = pyobject->ob_type->tp_name;
     }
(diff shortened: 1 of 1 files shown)


bsteffensmeier (approved):
Looks good to me, thanks

ndjensen (approved):

